PR #1, opened 2026-05-01
Fix Windows compatibility for resource import and UTF-8 encoding

irfansaf:
# Windows compatibility fixes

  `python app.py` doesn't start on Windows after the `task#1`/`task#2` series. A
  couple of POSIX-only imports need guarding before the app reaches Flask, plus
  a few text-mode `open()` calls decode under cp1252 by default and crash on the
  Indonesian content in the default plugin manifests.

 ## What's broken

  1. **Boot: `import resource`** — `routes/dashboard.py` imports `resource`
     (POSIX-only stdlib) at module level. `python app.py` fails with
     `ModuleNotFoundError: No module named 'resource'` before Flask initializes.

  2. **Boot: cp1252 decoding** — once `resource` is guarded, the next failure
     is `UnicodeDecodeError` reading `plugins/kanban/plugin.json` in
     `PluginManager()` and `SkillsManager()` constructors. Python on Windows
     defaults `open()` to `cp1252`, which can't decode the Indonesian text in
     default plugin/skill manifests.

  3. **Runtime: `/restart`** — the slash command handler and the super-agent
     restart tool both lazy-import `resource`. Same root cause, runtime path.

  4. **CLI: unused `import fcntl`** — `cli/commands.py` imports `fcntl` at the
     top but never calls it. The unused import alone breaks the CLI on Windows.

  5. **CLI: `signal.SIGKILL`** — doesn't exist on Windows; `stop_server()`
     crashes if SIGTERM doesn't take.

  6. **Tools: unencoded writes** — `patch`, `clear_log_file`, and `use_skill`
     have text-mode `open()` calls that crash on non-ASCII file content.

### DIFF
--- a/backend/plugin_lifecycle.py
+++ b/backend/plugin_lifecycle.py
@@ -101,7 +101,7 @@ class PluginManager:
             # Read manifest to know which events and slash commands this plugin subscribes to
             manifest_path = os.path.join(plugin_dir, 'plugin.json')
             if os.path.isfile(manifest_path):
-                with open(manifest_path) as f:
+                with open(manifest_path, encoding='utf-8') as f:
                     manifest = json.load(f)
                 events = manifest.get('events', [])
                 slash_commands = manifest.get('slash_commands', [])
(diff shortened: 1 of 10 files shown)


irfansaf:
Hey mas @anvie, I try to add support for windows. could u please check it?

PR #3, opened 2026-05-06
feat: add OpenCode Zen and OpenCode Go providers

riobahtiar:
- Two new entries in PROVIDER_DEFAULTS pointing at  https://opencode.ai/zen/v1 and https://opencode.ai/zen/go/v1 (both OpenAI-compatible, Bearer auth via OPENCODE_API_KEY)
- New `default_thinking` provider flag pre-enables thinking on newly created model rows. Set True for OpenCode Go because its catalog is dominated by always-thinking models (Kimi K2, DeepSeek V4, MiniMax M2) whose upstream rejects requests that omit reasoning_content on prior assistant messages. Applied only on creation — reconfigure preserves the user's manual thinking toggle.
- Add models.dev hint under the Model ID input, visible only when Type is Remote
- Mirror dropdown + hint changes across settings.html (live) and settings_models.html (orphan, kept consistent)
- Ignore AI agent local configs and editor metadata(.codex/, .opencode/, .claude/, .windsurf/, .codeium/, AGENTS.md, CLAUDE.md)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -296,3 +296,14 @@ keys/
 data/
 plan/
 run/
+
+# Ignore Other Agents settings and Editor configs
+.codex/
+.opencode/
+.claude/
+.windsurf/
+.windsurfrules
+.codeium/
+.codeiumignore
+AGENTS.md
+CLAUDE.md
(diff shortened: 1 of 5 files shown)


anvie:
Thanks mas!

Issue #5, opened 2026-05-07
need the windows installer


k1m0ch1:
installation one liner with windows are good option,
currently using windows,
any other agent have windows option, better have one.

k1m0ch1:
anyway working on this, I leave the issue to monitor the progress
<img width="869" height="517" alt="Image" src="https://github.com/user-attachments/assets/448edfcb-6437-4f9a-8c22-7a21a19cd0f2" />

k1m0ch1:
<img width="744" height="728" alt="Image" src="https://github.com/user-attachments/assets/ea9d4ecf-66ae-4fcc-8623-9f91a4e05c4b" />

k1m0ch1:
whenever installed need to delete $USER\.evonic
<img width="639" height="508" alt="Image" src="https://github.com/user-attachments/assets/480785dc-f09b-49f6-ba90-97d833b2cdc2" />

k1m0ch1:
cool its working now

<img width="1420" height="987" alt="Image" src="https://github.com/user-attachments/assets/5930c0f6-f479-4459-bf57-34e3e83b7f2d" />

k1m0ch1:
to start `.\evonic.bat` start

<img width="638" height="409" alt="Image" src="https://github.com/user-attachments/assets/bdf9ab92-b1b6-41c0-b1ea-4e73b082bfb9" />

k1m0ch1:
hi @anvie how to update the documentation at https://evonic.dev/getting-started/installation/ ?

k1m0ch1:
new PR https://github.com/anvie/evonic/pull/6

anvie:
Wow, thank you! That’s so much effort. 

I'm curious about how the agent works in Windows. Because the existing tools for the agent are `bash` and `runpy`, and `bash` doesn't run natively on Windows, is the agent using runpy?

k1m0ch1:
> Wow, thank you! That’s so much effort.
> 
> I'm curious about how the agent works in Windows. Because the existing tools for the agent are `bash` and `runpy`, and `bash` doesn't run natively on Windows, is the agent using runpy?

its not done yet I just park the PR first 😆 

still working with the agent tho

<img width="1331" height="689" alt="Image" src="https://github.com/user-attachments/assets/fc8f79de-8598-464f-b85b-ada0458f2b42" />

PR #6, opened 2026-05-07
add: windows installation support

k1m0ch1:
regarding with issue #5 

I just thinking it would be good to have windows installation support.

and its working perfectly on local

image

I don't see any documentation regarding https://evonic.dev/getting-started/installation/ 
I want to add the step to installation into that page.


anvie:
Untuk ini pending dulu ya, nunggu agent tools bisa confirmed jalan di Windows

PR #8, opened 2026-05-08
fix(llm_client): explicitly set stream: false in OpenAI-compatible payload

### LINKED ISSUE #7: fix: add explicit stream: false to prevent JSONDecodeError with 9router and streaming-default providers
## Description

When using evonic with [9router](https://github.com/decolua/9router) as an OpenAI-compatible LLM gateway, every LLM call fails with a cryptic error:

```
[INFO] [backend.agent_runtime.llm_loop] admin using model: Custom (claude-combo) (claude-combo)
[ERROR] [backend.agent_runtime.llm_loop] LLM error [unknown_error]: Expecting value: line 1 column 1 (char 0)
[INFO] [backend.agent_runtime.llm_loop] admin using model: Custom (claude-combo) (claude-combo)
[ERROR] [backend.agent_runtime.llm_loop] LLM error [unknown_error]: Expecting value: line 1 column 1 (char 0)
[INFO] [backend.agent_runtime.llm_loop] admin using model: Custom (claude-combo) (glm-combo)
[ERROR] [backend.agent_runtime.llm_loop] LLM error [unknown_error]: Expecting value: line 1 column 1 (char 0)
```

The error repeats on every message regardless of which model is selected through the router.

## Root Cause

`backend/llm_client.py` builds the OpenAI-compatible request payload without a `stream` field:

```python
payload = {
    "model": self.model,
    "messages": processed_messages,
    "max_tokens": max_tokens,
}
```

The [OpenAI Chat Completions specification](https://platform.openai.com/docs/api-reference/chat/create) defines `stream` as optional with default `false`. **Most providers honor this default** — OpenAI itself, Anthropic, OpenRouter, Together AI, Groq, DeepSeek, llama.cpp, and others all return a buffered JSON response when the field is omitted. This is why the issue has not surfaced with the more common providers.

9router takes a different approach. From its source (`open-sse/handlers/chatCore.js`):

```js
const clientRequestedStreaming = body.stream === true || ...;
let stream = providerRequiresStreaming ? true : (body.stream !== false);
```

When `body.stream` is `undefined`, the expression `undefined !== false` evaluates to `true`, so 9router selects streaming mode. The router then returns `text/event-stream` on HTTP 200. Evonic calls `response.json()` on that SSE body and crashes with `JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, which propagates up as `unknown_error`.

This is not a bug in 9router — it is a deliberate design choice. There is no server-side option to flip the default; the client must indicate its preference explicitly.

## Recommendation

Evonic does not currently support Server-Sent Events. There is no SSE parser, no delta accumulator, and no streaming pipeline in `llm_client.py`. Until streaming is implemented as a proper feature (separate, larger effort), the client must explicitly request buffered responses to match its actual capability.

The Ollama-format branch in the same file already does this correctly (line 412). The OpenAI-compatible branch should mirror that behavior.

## Proposed Fix

Add `"stream": False` to the OpenAI-compatible payload in `backend/llm_client.py`:

```python
payload = {
    "model": self.model,
    "messages": processed_messages,
    "max_tokens": max_tokens,
    "stream": False,
}
```

One-line change. Mirrors the Ollama path. No new dependencies, no schema changes, no UI changes.

## Why This Is Safe

- `stream` is a standard field in the OpenAI Chat Completions spec — no compliant provider will reject it.
- Providers that already default to `false` (the majority): no behavior change.
- Providers that default to streaming when the field is omitted (e.g. 9router): now correctly return buffered JSON.
- Evonic has no streaming infrastructure — explicit `false` is the truthful contract between client and server.
- Tightens spec compliance instead of relying on each provider's interpretation of an omitted field.

## Steps to Reproduce

1. Run a local 9router instance (default config).
2. In evonic Settings → Models, add a Custom model with `base_url = http://localhost:<port>/v1` and an API key issued by 9router.
3. Send any message to an agent using that model.
4. Observe `LLM error [unknown_error]: Expecting value: line 1 column 1 (char 0)` in the logs on every turn.

## Future Work (Out of Scope)

Adding real streaming support — SSE parser in `llm_client.py`, delta accumulation for `tool_calls` (which arrive as fragmented JSON strings keyed by `index`), progressive chunk delivery to the frontend, and a per-model streaming toggle in Settings → Models — would be a meaningful UX improvement (token-by-token rendering instead of buffered 10–30 second waits). It should be tracked as a separate feature issue, not bundled with this fix.

binsarjr:
## Summary

Closes #7

Adds `"stream": False` explicitly to the OpenAI-compatible chat completion payload in `backend/llm_client.py`. This prevents a `JSONDecodeError: Expecting value: line 1 column 1 (char 0)` crash that occurs when using evonic with providers or routers that default to streaming when the `stream` field is omitted (confirmed with [9router](https://github.com/decolua/9router)).

## Root Cause

The [OpenAI Chat Completions spec](https://platform.openai.com/docs/api-reference/chat/create) defines `stream` as optional with default `false`. Most providers honor this. However, 9router uses the following logic:

```js
let stream = providerRequiresStreaming ? true : (body.stream !== false);
```

When `body.stream` is `undefined` (field omitted), `undefined !== false` → `true` → router returns `text/event-stream`. Evonic calls `response.json()` on the SSE body and crashes.

## Changes

- `backend/llm_client.py` — add `"stream": False` to the OpenAI-compatible payload (1 line)

## Why This Is Safe

- `stream` is a standard OpenAI spec field — no compliant provider will reject it
- Providers that already default to `false`: no behavior change
- Providers that deviate from spec: now receive explicit instruction
- The Ollama-format path already sets `stream: False` — this mirrors that behavior
- Evonic has no SSE parser; explicit `false` matches actual capability

## Test Plan

- [ ] Start evonic with a 9router model configured (default 9router config)
- [ ] Send a message to an agent — confirm no `unknown_error` in logs
- [ ] Send a message using a standard OpenAI-compatible provider — confirm no regression

### DIFF
--- a/backend/llm_client.py
+++ b/backend/llm_client.py
@@ -423,6 +423,7 @@ class LLMClient:
                 "model": self.model,
                 "messages": processed_messages,
                 "max_tokens": max_tokens,
+                "stream": False,
             }
             if effective_temperature is not None:
                 payload["temperature"] = effective_temperature


anvie:
Thanks!

PR #15, opened 2026-05-09
fix(#10): release-mode path resolution & start parity

### LINKED ISSUE #10: Updating process error
I installed 0.1.8 then using this steps to setting up supervisor https://evonic.dev/agents/self-update/ for updating to 0.2.0 
but I got "new setup evonic" like installing a new system rather than updating.
fortunately this step `python3 supervisor/migrate.py --tag v0.2.0` copying* mutable data, so I got a "backup". 

since I'm not fluent of python nor linux/bash, but I saw something off.
while you installing with `curl -fsSL https://evonic.dev/install.sh | bash` it will setting up venv inside .evonic folder ( /home/user/.evonic/venv/)
but supervisor after updating process will run /home/user/.evonic/releases/v0.2.0/.venv/bin/python /home/user/.evonic/releases/v0.2.0/app.py
I don't know if this related because we got different virtual environment.
the confusing thing: if I typed `evonic start` it will run old installed evonic at `home/user/.evonic/` but when i typed `evonic start -d` it will run '/home/user/.evonic/releases/v0.2.0/.venv/bin/python /home/user/.evonic/releases/v0.2.0/app.py' and new setup page appeared.

currently I fresh installed v0.2.0 and copying (backuped) shared folder content into `home/user/.evonic/` then I re-setting-up self-update the confusing thing re-appeared. So right now I'm using "screen" to run evonic daemon-likely. 

*/ idk the docs said moving folder but in my case it copying

OS: Ubuntu 24
Maap bahasa inggrisnya jelek

binsarjr:
Fixes #10 — the "new setup" screen reappearing after self-update.

- `config.py`: resolve `APP_ROOT` (grandparent of `releases/<tag>/` in release mode, `BASE_DIR` otherwise) and route `DB_PATH` and log paths through it
- `cli/commands.py`: foreground `evonic start` detects release mode and execs into the release's python, matching `evonic start -d` behavior
- `supervisor/migrate.py` & `supervisor/supervisor.py`: detect install venv for `python_bin` instead of capturing system python; idempotent + safe `link_shared_dirs`; `start_daemon_from_current` relinks shared items before start

Tested end-to-end on a fresh v0.1.8 install: bug reproduces pre-fix, gone post-upgrade. 23 new unit tests, full suite passes (45/45).

See first comment for repro evidence and the few choices I made differently from the issue analysis.

### DIFF
--- a/cli/commands.py
+++ b/cli/commands.py
@@ -20,8 +20,22 @@ for lib_dir in ("lib",):
         sys.path.insert(0, lib_path)
 
 
-# PID file location
-PID_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", "run")
+# PID file location.
+#
+# After migration to the release-based layout the canonical run/ directory
+# lives at ``<app_root>/shared/run/`` (supervisor writes the daemon PID
+# there). Pre-migration installs still use ``<install_root>/run/``. Resolve
+# whichever is present so ``evonic status``/``evonic stop`` find a daemon
+# regardless of whether it was launched by the legacy in-process path or by
+# supervisor.
+def _resolve_pid_dir():
+    shared_run = os.path.join(ROOT, "shared", "run")
+    if os.path.isdir(shared_run):
+        return shared_run
+    return os.path.join(ROOT, "run")
+
+
+PID_DIR = _resolve_pid_dir()
 PID_FILE = os.path.join(PID_DIR, "evonic.pid")
 
 
--- a/cli/commands.py
+++ b/cli/commands.py
@@ -217,6 +217,29 @@ def start_server(port=None, host=None, debug=None, daemon=False):
     if debug:
         print("Debug mode: ON")
 
+    # Foreground release-mode parity: if the app was migrated to release-based
+    # layout, exec the release's python on its app.py (mirrors daemon path).
+    # Falls through to legacy in-process import when no release is staged.
+    current_link = os.path.join(ROOT, "current")
+    sup_cfg_path = os.path.join(ROOT, "supervisor", "config.json")
+    release_mode = os.path.islink(current_link) and os.path.exists(sup_cfg_path)
+    if release_mode:
+        release_path = os.path.realpath(current_link)
+        if sys.platform == "win32":
+            release_py = os.path.join(release_path, ".venv", "Scripts", "python.exe")
+        else:
+            release_py = os.path.join(release_path, ".venv", "bin", "python")
+        release_app = os.path.join(release_path, "app.py")
+        if os.path.exists(release_py) and os.path.exists(release_app):
+            print(EVONIC_BANNER)
+            print(f"Starting server (Ctrl+C to stop)")
+            print(f"Host: {host}")
+            print(f"Port: {port}")
+            print(f"URL: http://{host if host != '0.0.0.0' else 'localhost'}:{port}")
+            os.chdir(release_path)
+            os.execv(release_py, [release_py, release_app])
+            # execv replaces the process; lines below unreachable.
+
     try:
         from app import app
     except ModuleNotFoundError as e:
--- a/config.py
+++ b/config.py
@@ -60,7 +60,24 @@ def get_evaluator_type(domain: str) -> str:
 
 # Database paths
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
-_shared_db_dir = os.path.join(BASE_DIR, "shared", "db")
+
+
+def _resolve_app_root(base_dir: str) -> str:
+    """Return the app root directory.
+
+    In release mode the running code lives under ``<app_root>/releases/<tag>/``.
+    Mutable state (``shared/``, ``current`` symlink) lives at the app root, so
+    config must resolve to the grandparent in that case. In flat-repo mode the
+    app root *is* the directory containing this file.
+    """
+    parent = os.path.dirname(base_dir)
+    if os.path.basename(parent) == "releases":
+        return os.path.dirname(parent)
+    return base_dir
+
+
+APP_ROOT = _resolve_app_root(BASE_DIR)
+_shared_db_dir = os.path.join(APP_ROOT, "shared", "db")
 if not os.path.isdir(_shared_db_dir):
     os.makedirs(_shared_db_dir, exist_ok=True)
 
@@ -84,10 +101,10 @@ LOG_FULL_RESPONSE = os.getenv("LOG_FULL_RESPONSE", "0") == "1"
 
 # Raw LLM API call logging (markdown)
 LLM_API_LOG_ENABLED = os.getenv("LLM_API_LOG_ENABLED", "0") == "1"
-LLM_API_LOG_FILE = os.getenv("LLM_API_LOG_FILE", os.path.join(BASE_DIR, "logs", "llm_api_calls.md"))
+LLM_API_LOG_FILE = os.getenv("LLM_API_LOG_FILE", os.path.join(APP_ROOT, "logs", "llm_api_calls.md"))
 
 # Event stream logging to file
-EVENT_LOG_FILE = os.getenv("EVENT_LOG_FILE", os.path.join(BASE_DIR, "logs", "events.log"))
+EVENT_LOG_FILE = os.getenv("EVENT_LOG_FILE", os.path.join(APP_ROOT, "logs", "events.log"))
 
 # Docker sandbox configuration (shared by runpy, bash, etc.)
 SANDBOX_WORKSPACE = os.getenv("SANDBOX_WORKSPACE", BASE_DIR)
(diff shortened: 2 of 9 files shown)


binsarjr:
Some extra context for reviewers — repro evidence, design choices, and one known gap.

## End-to-end repro

Fresh v0.1.8 install on macOS, ran the setup wizard, then `supervisor/migrate.py` to switch to release-based layout. Started the daemon:

```
$ evonic start -d
$ curl -I http://localhost:28080/
HTTP/1.1 302 Found
Location: http://localhost:28080/setup    ← bug
```

Probed config from inside the release process:

```
BASE_DIR  = .../releases/v0.1.8
DB_PATH   = .../releases/v0.1.8/shared/db/evonic.db
DB exists = True   (282KB, schema only — created by os.makedirs + sqlite init)
has_super = False
```

The real DB at `<app_root>/shared/db/evonic.db` (282KB with the actual super agent) was never read.

Staged a release from this branch, atomic-swapped `current`, restarted:

```
BASE_DIR  = .../releases/v0.2.0-fix
APP_ROOT  = .../              ← grandparent of releases/
DB_PATH   = .../shared/db/evonic.db
DB exists = True
has_super = True
```

```
$ curl -I http://localhost:28080/
HTTP/1.1 302 Found
Location: http://localhost:28080/login    ← fixed
```

## Design choices that diverged from #10's analysis

**Relpath instead of abspath in `link_shared_dirs`.** Issue suggested switching supervisor to abspath to match migrate. I went the other direction — aligned migrate to relpath. Couldn't reproduce the "atomic_swap breaks relative symlinks" case (swap only repoints `current`, the release dir stays put). Relpath also survives `mv ~/.evonic somewhere-else`. Happy to flip if there's a production scenario I'm missing.

**Defensive guard in `link_shared_dirs`.** Old code did `shutil.rmtree(link)` if a real dir sat at the link path. I made it refuse and log instead — the self-update flow runs unattended, and nuking a manually-recovered directory felt scary. Refusal is item-scoped, other links still proceed.

**Idempotency.** The function now skips if the symlink already resolves to the right target. Needed this so `start_daemon_from_current` can call it on every restart without churn.

## Known gap (not blocking)

Foreground `evonic start` execs into the release's python — but only if the install-root `cli/commands.py` has this fix. The current self-update flow swaps `current` but doesn't update install-root scripts, so existing installs will keep using their old foreground CLI until they re-install. Daemon mode (the actual issue here) is unaffected since supervisor invokes the release's python directly.

Probably worth a separate issue for "self-update should also refresh install-root entry points" — happy to file if it sounds right.

binsarjr:
Pushed a follow-up commit for the PID issue @vachzar surfaced. Tl;dr: same root cause as before, just a different symptom.

## Why status/stop were failing

`supervisor.start_daemon` spawns the release process and returns the PID via the function tuple — but never persists it to disk. Pre-migration that was fine because the legacy in-process path in `cli/commands.py start_server` was the one writing `evonic.pid` (the cli's daemon mode wrote PID after `subprocess.Popen`). Once supervisor became the primary daemon launcher in release mode, that write was simply gone.

On top of that, the cli and supervisor were already disagreeing about *where* the PID lives:

- cli read/write: `<install_root>/run/evonic.pid`
- supervisor's `_pid_file()`: `<app_root>/shared/run/evonic.pid`

In release mode they didn't even point at the same file (no `run` symlink at install root post-migrate). So even if supervisor *had* started writing, the cli was looking somewhere else.

## What this commit changes

- `supervisor.start_daemon` writes `proc.pid` to `<app_root>/shared/run/evonic.pid` after the readiness check passes
- `supervisor.stop_daemon` removes the pid file on every exit path (already-dead, graceful, post-SIGKILL)
- `cli/commands.py` resolves `PID_FILE` to `shared/run/evonic.pid` when the shared layout exists, else falls back to the legacy `run/evonic.pid` for pre-migration installs

Both sides now point at the same file, and supervisor actually populates it.

## Verification

Same sandbox as before — install at v0.1.8, set up, migrated, upgraded to this branch, then:

```
$ evonic start -d
Supervisor started (PID: 19891)
Server akan berjalan dari current release dengan self-update otomatis

$ cat shared/run/evonic.pid
19892

$ evonic status
Server is running (PID: 19892)

$ evonic stop
Sending SIGTERM to server (PID: 19892)...
Server didn't stop gracefully. Force-killing...
Server killed (PID: 19892)
Sending SIGTERM to supervisor (PID: 19891)...

$ ls shared/run/
server.log
```

Both daemon and supervisor stopped, PID file cleaned up.

## Tests

Added 9 tests in `unit_tests/test_supervisor/test_pid_persist.py` covering write/remove helpers, `start_daemon`-writes-pid (success and early-exit cases), and `stop_daemon`-removes-pid (alive, dead, missing). 54 supervisor tests pass total.

## Note on the foreground gap from earlier

Same caveat applies — the PID write only kicks in when install-root `supervisor.py` has this code. Existing installs running on old install-root supervisor will still miss the write until they re-install or until a future change refreshes install-root entry points during self-update. Daemon mode through this PR is fully self-consistent once the release containing it is current.

anvie:
Thanks for the PR, the PID persistence work is solid, but there are three things that need fixing before merge:

1. _resolve_app_root heuristic in config.py - string-matching on directory name "releases" is not how you detect a release layout. That breaks the moment someone names a directory differently. Use a marker file like current symlink, shared/ directory, or a VERSION file. Anything that actually proves the layout, instead of guessing from a name.
2. detect_python_bin() duplication - it's copy-pasted verbatim across supervisor/supervisor.py and supervisor/migrate.py with inconsistent platform checks (is_windows() in one, sys.platform == 'win32' in the other). Consolidate into one shared location. I don't care where, just don't make me look at two copies of the same function and wonder which one has the bugs.
3. _remove_daemon_pid TOCTOU - you removed the PID file in all five exit paths, which is great, but wrap that os.remove in try/except FileNotFoundError. Right now a race between checking existence and deleting is a theoretical crash waiting to happen. Not theoretical enough to ignore.

Rest looks good 👍

binsarjr:
Thanks for taking the time on this — points 2 and 3 were good catches, both pushed.

On point 1, let me push back a little (open to being wrong here): I treated `releases/` as a stable structural name because the rest of the supervisor/migrate code already hardcodes it — 17 path-construction references on `main` today. A few representative spots:

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/supervisor/supervisor.py#L158-L167

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/supervisor/supervisor.py#L745-L750

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/supervisor/migrate.py#L83-L90

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/supervisor/migrate.py#L173-L178

If a user renamed `releases/` to anything else, the whole self-update flow would break before `_resolve_app_root` even runs — `get_current_release` wouldn't find the tag, `atomic_swap` would write to the wrong place, etc. So in practice the directory name is a convention the rest of the code already enforces.

---

While auditing for this, I found the convention is actually inconsistent across the codebase. There are **three** detection points that depend on the literal `'releases'` string:

1. `config.py` — the heuristic you flagged.
2. `backend/slash_commands.py` (`/restart` command):

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/backend/slash_commands.py#L335-L350

3. `backend/tools/super_agent_tools.py` (super-agent restart tool):

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/backend/tools/super_agent_tools.py#L696-L710

And **four** other detection points that are already marker-based (no name dependency):

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/cli/commands.py#L145-L155

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/plugins/kanban/db.py#L18-L21

https://github.com/anvie/evonic/blob/e5df2920d7f7d0c1ea8133ed80798f710e110959/plugins/agentapi/db.py#L18-L21

So the codebase is already split between two styles, and your point applies equally to all three name-based detections — fixing only `config.py` would leave `/restart` and the super-agent tool with the same fragility.

A few ways to land this:

**Option A — fix all three in this PR.** Extract a small `is_release_mode(path)` helper (checks `current` symlink + `shared/` directory at the candidate app_root), use it from `config._resolve_app_root`, `slash_commands.py`, and `super_agent_tools.py`. Keeps scope creep contained but self-contained — one concept, one helper, three callers. Adds maybe ~20 LOC.

**Option B — separate issue/PR.** Land points 2 and 3 here, file a follow-up issue titled something like "Replace name-based release-mode detection with structural markers" covering all three sites + a shared helper. Keeps this PR focused on issue #10's original scope; the structural-detection work becomes its own reviewable change.

I lean slightly towards B because (a) it's a wider design change than issue #10 was scoped for, (b) it touches restart paths that need their own test/manual-verify pass, and (c) it gives space to discuss whether the "principle" actually buys anything practical given the 17 path-construction references would still cement the name. But A is also reasonable if you'd rather not multiply tickets.

Either way, want to flag this so we don't accidentally lock in a half-converted state. WDYT? Or am I still misreading the flow somewhere?

PR #25, opened 2026-05-13
feat(update): persist update progress to survive crashes

DeryFerd:
## PR Summary

Persists update progress to disk so state survives server crashes and restarts, enabling better debugging and recovery from failed updates.

## Problem

When the server crashes during an update (power loss, OOM kill, manual restart), all progress information is lost. Users don't know:
- What step the update failed at
- What error occurred
- Whether the update was even in progress

This makes debugging and recovery difficult.

## Solution

Implemented atomic state persistence to `shared/update/update_state.json`. The update manager now:
1. Saves state after every progress update
2. Detects crashed updates on startup
3. Preserves logs and error messages across restarts

### State Persistence

- **Storage Location**: `shared/update/update_state.json`
- **Write Strategy**: Atomic (write to `.tmp`, then `os.replace()`)
- **Persistence Triggers**: After every log entry, progress update, and status change

### Crash Detection

On module import (server startup):
- Loads persisted state from disk
- Detects if previous update was interrupted (`status == 'updating'`)
- Automatically marks as failed with descriptive error
- Logs warning with last known step and target version

### New Status Fields

- `crashed`: Boolean flag indicating detected crash (cleared after first read)
- `last_update_attempt`: Unix timestamp of last update start

### Updated Functions

All state-modifying functions now persist to disk:
- `check_for_update()` - Persists check results
- `start_update()` - Persists initial state
- `WebNotifier.send_progress()` - Persists after each step
- `WebNotifier.send_failure()` - Persists failure state
- `WebNotifier.send_success()` - Persists success state
- `trigger_rollback()` - Persists rollback state

## Example Recovery Flow

**Before** (crash during update):
```
[Server crashes at step 4/6]
[Server restarts]
User: "What happened? Where's my update?"
Status: idle, no information
```

**After** (with persistence):
```
[Server crashes at step 4/6]
[Server restarts]
Log: "Detected incomplete update to v0.2.7 (was at step 4/6)"
Status: failed
Error: "Update interrupted (server crash or restart)"
Logs: Full history preserved
```

## Testing

- ✅ Syntax validation passed
- ✅ Atomic write logic verified
- ✅ No breaking changes to existing API

## Files Changed

- `backend/update_manager.py`: Added persistence layer and crash detection

## Risk Assessment

**Risk Level**: Low

- State file is in `shared/` directory (survives updates)
- Atomic writes prevent corruption
- Graceful fallback if state file is corrupted
- No changes to update execution logic

## Benefits

- **Better Debugging**: Full context available after crashes
- **User Transparency**: Clear indication of what went wrong
- **Recovery Information**: Logs and error messages preserved
- **Operational Insight**: Track update attempts and failure patterns

## Future Enhancements

This foundation enables:
- Update retry logic with backoff
- Automatic rollback on crash detection
- Update analytics and failure tracking
- Web UI recovery suggestions

### DIFF
--- a/backend/update_manager.py
+++ b/backend/update_manager.py
@@ -3,6 +3,8 @@ Server-side update state manager.
 
 Provides daily-cached update checks, background update execution with log
 capture, and SSE listener management for real-time web UI notifications.
+
+Progress state is persisted to disk to survive crashes and restarts.
 """
 
 import json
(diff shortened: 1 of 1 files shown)


anvie:
I see that TOTAL_STEPS doesn’t exist anywhere, the "TOTAL_STEPS" in dir() check is always False, so crash logs always show /6 regardless of actual step count.

_state mutated & persisted outside _lock in multiple places, can be potentially data race.

Double _persist_state in send_progress/send_failure/send_success, _append_log already persists, doubling disk I/O on every step.

DeryFerd:
Hi @anvie, thanks for the detailed review! I've addressed all three issues:

1. TOTAL_STEPS undefined - Added TOTAL_STEPS = 6 constant at module level, now crash logs show correct step count
2. Data race - Wrapped all _state mutations with _lock in: WebNotifier.begin(), send_progress(), send_failure(), send_success(), check_for_update(), start_update(), _run_update_thread(), trigger_rollback(). Crash detection now also uses _lock and persists properly
3. Double persistence - Removed duplicate _persist_state() calls from WebNotifier methods, now only _append_log() persists (single I/O per operation)

All CI checks are passing. Ready for re-review!

PR #26, opened 2026-05-14
security(tools): protect .env files in write operations

DeryFerd:
This PR adds protection for `.env` files across all file operation tools. Right now, agents with file access can read or modify environment files that typically contain API keys, database passwords, and other secrets.

### Changes

The `safety_checker.py` already had a `check_env_path()` function that detects `.env` files and their variants (`.env.local`, `.env.production`, etc.), but only `read_file.py` was actually calling it. This PR adds the same check to the three other file operation tools that were missing it:

- `write_file.py`
- `patch.py` 
- `str_replace.py`

When a regular agent tries to access an `.env` file through any of these tools, the system now requires user approval before proceeding. Super agents can still bypass the check when needed.

### Why?

Environment files are a common target in AI agent attacks. An agent that can read `.env` gets immediate access to every secret in your application. An agent that can write to `.env` can inject malicious credentials or overwrite legitimate ones.

The protection was already there for reading, but writing and patching were unguarded. This closes that gap.

### Testing

Added 21 unit tests covering:
- All four file operation tools (read, write, patch, str_replace)
- Multiple `.env` variants (`.env`, `.env.local`, `.env.production`, `.env.test`, etc.)
- Super agent bypass behavior
- Edge cases (subdirectories, files named "env" without the dot, safety checker disabled)

All tests pass:
```bash
$ python -m pytest unit_tests/test_env_file_protection.py -v
21 passed
```

### Compatibility

This change is backward compatible. Existing agents continue to work exactly as before. The only difference is that operations on `.env` files now trigger the approval flow instead of executing silently.


anvie:
Thank you!

PR #32, opened 2026-05-15
fix(app): limit unreplied-chat startup scan to human-facing sessions

DeryFerd:
## Summary

On startup, Evonic scans chat sessions for user messages that never got an agent reply (for example after a restart or deploy). That scan was treating **every** session the same, including agent-to-agent, scheduler, and system notification traffic. Those are not real “user waiting for a reply” cases, so they could produce noisy or misleading warnings in the logs.

This PR tightens the scan so it only considers **human-facing** sessions: web UI chats and channel users (Telegram, WhatsApp, etc.).

## Motivation

There was an open TODO in `app.py` asking to limit the unreplied-chat check to true human↔agent conversations. The previous filter only skipped `__agent__*` and `__scheduler__`, but still included:

- `__system__` / `__system__<agent_id>` (internal notifications)
- Empty or invalid `external_user_id` values
- Sub-agent rows if they appear in the enabled agent list

That made the startup warning less trustworthy for operators debugging real user issues.

## What changed

- Added `is_human_facing_external_user_id()` in `models/chat.py` as a single source of truth for “is this a human session?”
- Updated `_check_unreplied_chats()` in `app.py` to:
  - Skip non-human sessions via the helper
  - Skip `is_subagent` agents
  - Re-enable the end-of-scan summary log (now labeled “human session(s)”)
- Added `unit_tests/test_human_facing_session.py` with 9 cases covering web users, channels, agent/system IDs, and edge cases

## Behavior

| Session type | Scanned? |
|--------------|----------|
| Web UI user (`user_123`, `web_test`) | Yes |
| Channel user (e.g. Telegram ID) | Yes |
| `__agent__*` (agent-to-agent) | No |
| `__scheduler__` | No |
| `__system__` / `__system__*` | No |
| Empty `external_user_id` | No |

`web_test` is intentionally **included** — it represents browser testing in the web UI, consistent with `get_web_fallback_session()`.

## Testing

```bash
python -m pytest unit_tests/test_human_facing_session.py -v
```

All 9 tests pass.

No manual app run required for this change; it is log-only at startup and covered by unit tests. Optional sanity check: start Evonic with a mix of human Telegram/web sessions and an active __agent__ session — only human sessions should appear in unreplied-chat warnings.

Risk
Low. The scan is diagnostic logging only; it does not change message routing, agent replies, or session creation.

### DIFF
--- a/app.py
+++ b/app.py
@@ -260,16 +260,18 @@ if not _reloader_active or _is_reloader_child:
     # Startup check: scan all active sessions for unreplied user messages
     # ----------------------------------------------------------------
     def _check_unreplied_chats():
-        """Scan all chat sessions on startup. Log any session where the last
-        message is from a user and no agent has replied — these users may have
-        been left hanging after a restart or deployment."""
-        
-        # @TODO(robin): Perlu pastikan bahwa chat session-nya hanya yg user sama agent saja, jangan scan session yg dari agent-to-agent atau system-to-agent. contoh user to agent itu seperti chat di halaman agent detail chat tab, atau chat melalui channel seperti telegram.
+        """Scan human-facing chat sessions on startup. Log any session where the
+        last message is from a user and no agent has replied — these users may
+        have been left hanging after a restart or deployment.
 
+        Only scans sessions between a human and an agent (web UI or channels).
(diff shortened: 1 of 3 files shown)


anvie:
Thank you!

PR #36, opened 2026-05-15
fix(workplaces): delegate file I/O on LocalWorkplaceBackend

### LINKED ISSUE #34: LocalWorkplaceBackend crashes on every bash/runpy call - v0.3.19
## What happens

When I try to run any bash or runpy command from an agent, I get this error right away:

```
TypeError: Can't instantiate abstract class LocalWorkplaceBackend
without an implementation for abstract methods
'file_exists', 'file_stat', 'make_dirs', 'read_file', 'write_file'
```

Interesting thing, the file tools like `read_file` and `write_file` still work fine. Only `bash` and `runpy` are affected.

## What I found when looking at the code

I tried to understand the error by reading the source. I am not a Python expert, but here is what I see:

The file `backend/workplaces/backends/local_workplace.py` has a class called `LocalWorkplaceBackend`. It only has these methods inside: `__init__`, `_get_inner`, `run_bash`, `run_python`, `destroy`, `status`.

But its parent class `ExecutionBackend` (in `backend/tools/lib/exec_backend.py`) lists 5 more methods with `@abstractmethod` on top: `file_exists`, `file_stat`, `read_file`, `write_file`, `make_dirs`.

From what I read, Python requires the child class to implement all abstract methods from the parent. Since these 5 are missing, Python refuses to create the object, and that is the TypeError.

I also compared this file between v0.2.6 and v0.3.19. The two files are exactly the same (byte for byte). So this bug has been there for a while.

## What I tried, and it worked

I added 5 simple methods to `LocalWorkplaceBackend`, right after `status()`:

```python
def file_exists(self, path: str) -> bool:
    return self._get_inner().file_exists(path)

def file_stat(self, path: str) -> dict:
    return self._get_inner().file_stat(path)

def read_file(self, path: str) -> dict:
    return self._get_inner().read_file(path)

def write_file(self, path: str, content: str, create_dirs: bool = True) -> dict:
    return self._get_inner().write_file(path, content, create_dirs=create_dirs)

def make_dirs(self, path: str) -> dict:
    return self._get_inner().make_dirs(path)
```

Each one just forwards the call to `_get_inner()`. I checked, both `LocalBackend` and `DockerBackend` already have these methods with the same arguments, so I think this is safe.

So for now I am running with this local patch, which means I am not blocked. I am filing this issue so you can decide the proper fix for the next release.

## How to see the bug

1. Fresh install with `curl -fsSL https://evonic.dev/install.sh | bash`
2. Make a Local workplace pointing to any folder
3. Make an agent with `bash` and `runpy` tools
4. In the agent chat, try `pwd` or any other shell command
5. The TypeError shows up

DeryFerd:
## Background

[#34](https://github.com/anvie/evonic/issues/34) reports that any `bash` or `runpy` call from an agent bound to a **local workplace** dies immediately with:

```text
TypeError: Can't instantiate abstract class LocalWorkplaceBackend
without an implementation for abstract methods
'file_exists', 'file_stat', 'make_dirs', 'read_file', 'write_file'
```

File tools (`read_file`, `write_file`, etc.) can still work because they often go through other backends. Shell and Python execution go through `WorkplaceManager` → `LocalWorkplaceBackend`, so the failure is total for those tools.

The class in `backend/workplaces/backends/local_workplace.py` has looked the same since v0.2.6. What changed is that `ExecutionBackend` in `backend/tools/lib/exec_backend.py` now declares five file I/O methods as `@abstractmethod` so tools like `write_file` and `patch` target the same environment as `bash`/`runpy` (Docker, SSH, workplace, etc.). `LocalWorkplaceBackend` never caught up, so Python refuses to construct it.

## What this change does

Add five one-line delegators on `LocalWorkplaceBackend`, each forwarding to `_get_inner()`:

- `file_exists`
- `file_stat`
- `read_file`
- `write_file` (including `create_dirs`)
- `make_dirs`

`_get_inner()` already returns either `LocalBackend` or `DockerBackend` (sandbox path). Both inner classes implement the full interface, so there is no new logic, only wiring.

No change to workplace config, session routing, or sandbox defaults.

## Files touched

| File | Change |
|------|--------|
| `backend/workplaces/backends/local_workplace.py` | Implement the five abstract methods |
| `unit_tests/test_local_workplace_backend.py` | New tests |

## Tests

`unit_tests/test_local_workplace_backend.py`:

- Instantiates `LocalWorkplaceBackend` and checks it is a concrete `ExecutionBackend` (would have failed before on 3.12+ abstract base enforcement).
- Mocks the inner backend and asserts each file method calls through with the expected arguments.

## How a reviewer can sanity-check

1. Create a local workplace pointing at any directory.
2. Attach an agent with `bash` and/or `runpy`.
3. Run something trivial (`pwd`, `print(1)`).

Before: TypeError on first tool call. After: command runs in the workplace workspace.

I did not run a full fresh install in this PR; validation is unit tests plus the code path review above. Happy to run a manual pass if you want that noted in the PR thread.

## Risk / compatibility

Low. The methods are pure delegation; behavior matches whatever `LocalBackend` or `DockerBackend` already does for non-workplace sessions.

Sandbox workplaces (`sandbox_enabled=True`) lazy-init `DockerBackend` on first use; file calls after that follow the same inner instance as `run_bash`.

Closes [#34](https://github.com/anvie/evonic/issues/34).

### DIFF
--- a/backend/workplaces/backends/local_workplace.py
+++ b/backend/workplaces/backends/local_workplace.py
@@ -54,3 +54,18 @@ class LocalWorkplaceBackend(ExecutionBackend):
             'workspace': self._workspace,
             'status': 'idle',
         }
+
+    def file_exists(self, path: str) -> bool:
+        return self._get_inner().file_exists(path)
+
+    def file_stat(self, path: str) -> dict:
+        return self._get_inner().file_stat(path)
+
+    def read_file(self, path: str) -> dict:
+        return self._get_inner().read_file(path)
+
+    def write_file(self, path: str, content: str, create_dirs: bool = True) -> dict:
+        return self._get_inner().write_file(path, content, create_dirs=create_dirs)
+
+    def make_dirs(self, path: str) -> dict:
+        return self._get_inner().make_dirs(path)
--- a/unit_tests/test_local_workplace_backend.py
+++ b/unit_tests/test_local_workplace_backend.py
@@ -1,0 +1,45 @@
+"""Tests for LocalWorkplaceBackend ExecutionBackend compliance."""
+
+import unittest
+from unittest import mock
+
+from backend.tools.lib.exec_backend import ExecutionBackend
+from backend.workplaces.backends.local_workplace import LocalWorkplaceBackend
+
+
+class TestLocalWorkplaceBackend(unittest.TestCase):
+    def test_is_concrete_execution_backend(self):
+        backend = LocalWorkplaceBackend(
+            config={'workspace_path': '/tmp/ws'},
+            sandbox_enabled=False,
+        )
+        self.assertIsInstance(backend, ExecutionBackend)
+
+    def test_file_methods_delegate_to_inner(self):
+        inner = mock.MagicMock()
+        inner.file_exists.return_value = True
+        inner.file_stat.return_value = {'exists': True, 'size': 1, 'is_binary': False}
+        inner.read_file.return_value = {'content': 'hi'}
+        inner.write_file.return_value = {'ok': True}
+        inner.make_dirs.return_value = {'ok': True}
+
+        backend = LocalWorkplaceBackend(
+            config={'workspace_path': '/tmp/ws'},
+            sandbox_enabled=False,
+        )
+        backend._inner = inner
+
+        self.assertTrue(backend.file_exists('/a'))
+        inner.file_exists.assert_called_once_with('/a')
+
+        self.assertEqual(backend.file_stat('/a')['size'], 1)
+        inner.file_stat.assert_called_once_with('/a')
+
+        self.assertEqual(backend.read_file('/a')['content'], 'hi')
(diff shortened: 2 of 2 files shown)


anvie:
Thank you for the PR! However, this fix was already applied
directly to main in commit bc236377fc9152c715429a887ffc442259275a3d
by anvie on May 15, 2026, which closed issue #34.

LocalWorkplaceBackend already has file_exists, file_stat, read_file,
write_file, and make_dirs — all delegating to _get_inner() — exactly
the same pattern this PR proposes.

The PR branch now conflicts with main since both sides added the
same methods at different positions in the file.

Closing as the fix is already in main. The unit test file
(unit_tests/test_local_workplace_backend.py) is the one piece not
in main yet — worth cherry-picking if you want test coverage for
the delegation.

Best,
Linus T.
--
Robin Syihab's agent.

PR #39, opened 2026-05-16
security(supervisor): re-enable release tag signature verification

DeryFerd:
## Summary

Release updates were skipping tag signature checks entirely. In `supervisor/supervisor.py`, step 2 of `run_update()` had verification gated behind `if False`, so every non-nightly update continued even when `git verify-tag` would have failed.

This PR turns that check back on for normal tagged releases. When `skip_verify` is false (the default), the supervisor calls `verify_tag()` and aborts the update with `UpdateError` if GPG/SSH verification fails.

Intentional bypass paths stay the same:

- **Nightly builds** still skip verification (unchanged).
- **`skip_verify=True`** still skips it — this is what `evonic update --force` uses for local/dev installs without signed tags.

Plugin zip signing (separate PR) covers install-time artifacts; this change is specifically about **release tags** fetched during `evonic update`.

## What changed

- `supervisor/supervisor.py`: restore `if not skip_verify:` around `verify_tag()`; clearer log/notify text when verification runs vs when `--force` skips it.
- `unit_tests/test_supervisor/test_lifecycle.py`: happy path now expects `verify_tag` to run; added coverage for `skip_verify=True`.

## Risk / compatibility

- Production installs that pull **unsigned** tags from `origin` will see updates fail until tags are signed or the operator uses `--force`. That matches the intended supply-chain posture for tagged releases.
- No change to nightly update flow or to CLI flags beyond existing `--force` behavior.

## Validation

- `python -m pytest unit_tests/test_supervisor/ -q`
- Not run: full `evonic update` against a live signed tag (would need a staged install with GPG/SSH trust configured).

### DIFF
--- a/supervisor/supervisor.py
+++ b/supervisor/supervisor.py
@@ -1070,15 +1070,14 @@ def run_update(tag: str, cfg: dict, notifier: Optional[TelegramNotifier],
             _notify(notifier, step, 'Skipping signature verification (nightly)')
             log.warning('Signature verification SKIPPED (nightly)')
         else:
-            # TODO: Re-enable signature verification for production
-            # Currently disabled for development convenience
-            _notify(notifier, step, 'Skipping signature verification (dev mode)')
-            if False:  # was: if not skip_verify:
+            if not skip_verify:
+                _notify(notifier, step, f'Verifying signature for {tag}')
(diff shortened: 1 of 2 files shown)


DeryFerd:

Issue #35, opened 2026-05-15
Docker Sandbox toggle in agent settings has no effect


husnimubarok:
## What happens

I turned on the "Docker Sandbox" toggle in my agent settings, but nothing changed. The agent still runs bash commands on my Mac (showing my username `husnix` and macOS paths), not inside a Docker container.

I also checked `docker ps -a` after the agent ran a command. No container is created, no matter if the toggle is on or off.

## What I found in the code

In `backend/workplaces/manager.py` line 225 I see this:

```python
def _create_local(self, config: dict) -> ExecutionBackend:
    from backend.workplaces.backends.local_workplace import LocalWorkplaceBackend
    return LocalWorkplaceBackend(config=config, sandbox_enabled=False)
```

The value `sandbox_enabled=False` is written directly in the code. So no matter what the agent toggle says in the UI, this code always sends `False`. The toggle setting from the database is never read here, as far as I can tell.

Note: this bug is hidden behind another crash from #34. You can only see this second bug after fixing that crash first.

## What I tried

For testing, I just changed `False` to `True`:

```python
return LocalWorkplaceBackend(config=config, sandbox_enabled=True)
```

After this, a Docker container actually starts with the name pattern `evonic-workplace-local-...`, my agent runs commands inside Debian 12, and `/workspace` is mounted correctly. So the Docker code does work, it just was never called.

But this is not a real fix because now every Local workplace is forced into Docker mode. The user cannot choose anymore.

For now I am running with the `True` change as a local patch, so my setup works. I am filing this so you can pick the proper way to wire the toggle in the next release.

anvie:
Closing — this has been fixed in commit 7c07724 (fix(#240): wire sandbox_enabled through WorkplaceManager for local workplaces).

PR #41, opened 2026-05-16
fix(agents): enforce sandbox disabled on remote and cloud workplaces

DeryFerd:
## Background

Docker sandbox mode only applies to **local** workplaces. Remote SSH and cloud (Evonet) workplaces run tools through different execution backends, so a sandbox toggle on those agents does not mean what the UI implies.

There was a gap in how agent create/update handled `sandbox_enabled`:

- On **create**, validation only ran when `sandbox_enabled` was truthy, so sending `0` skipped the workplace check entirely.
- On **update**, the same truthy guard applied, and when the user tried to enable sandbox on a remote/cloud workplace the code **deleted** `sandbox_enabled` from the payload instead of writing `0`. The old value in SQLite could stay out of sync with what the settings page shows.

Issue [#35](https://github.com/anvie/evonic/issues/35) reported the toggle feeling ineffective; the underlying issue is that policy was not applied consistently on every save path.

## How it behaves now

A small helper, `_apply_sandbox_workplace_policy()`, looks up the target workplace and, for `remote` or `cloud` types, always sets `sandbox_enabled` to `0` on the data dict passed to the database layer.

Both API entry points call it:

- **POST** `/api/agents` — uses `workplace_id` from the create body.
- **PUT** `/api/agents/<id>` — uses `workplace_id` from the body when present, otherwise the agent’s existing workplace.

That means a partial update (name, prompt, channel, etc.) on a cloud agent still persists `sandbox_enabled=0`, even when the client omits the sandbox field. Local workplaces are unchanged: both `0` and `1` pass through as sent.

The agent detail UI already disables the sandbox toggle for remote/cloud; this change makes the stored agent record match that expectation.

## Files touched

- `routes/agents.py` — `_apply_sandbox_workplace_policy()` plus calls from create/update handlers.
- `unit_tests/test_agent_sandbox_workplace_validation.py` — unit tests against the helper (local allowed, remote/cloud forced off, no lookup when `workplace_id` is missing).

## Notes for reviewers

- This is separate from workplace file I/O delegation ([#36](https://github.com/anvie/evonic/pull/36)).
- No migration: agents with stale values get corrected the next time they are updated through the API; new creates are correct immediately.
- Setup wizard / CLI agent flows were not modified in this PR; happy to follow up if those paths need the same helper.

## Checked with

- `python -m pytest unit_tests/test_agent_sandbox_workplace_validation.py -q`

### DIFF
--- a/routes/agents.py
+++ b/routes/agents.py
@@ -7,7 +7,7 @@ import re
 import json
 import uuid
 import queue
-from typing import Dict, Any, List
+from typing import Dict, Any, List, Optional
 from flask import Blueprint, render_template, jsonify, request, Response, stream_with_context
 from models.db import db
 from models.chatlog import chatlog_manager, _DISPLAY_TYPES
@@ -29,6 +29,15 @@ def _sanitize_agents(agents: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
     for a in agents:
         _sanitize_agent(a)
     return agents
+
+
+def _apply_sandbox_workplace_policy(agent_data: dict, workplace_id: Optional[str]) -> None:
+    """Docker sandbox is only supported on local workplaces."""
+    if not workplace_id:
+        return
+    workplace = db.get_workplace(workplace_id)
+    if workplace and workplace.get('type') in ('remote', 'cloud'):
+        agent_data['sandbox_enabled'] = 0
 
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 AGENTS_DIR = os.path.join(BASE_DIR, 'agents')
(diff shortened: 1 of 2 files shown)


anvie (approved):
Thank you for the PR! This is a clean fix for a real bug (issue #35).

The two bugs in the old code are clear:
1. **Create**: `if data.get('sandbox_enabled')` skips the check entirely when the client sends `sandbox_enabled=0` — a remote/cloud agent could be created with stale sandbox state.
2. **Update**: same falsy guard *plus* `del data['sandbox_enabled']` instead of setting to `0` — the old DB value persists through partial updates.

The new `_apply_sandbox_workplace_policy()` helper fixes both: always called, always writes `0` for remote/cloud.

Test coverage is solid — all three workplace types, both on/off states, and the no-workplace-id edge case with `assert_not_called`. Clean use of `unittest.mock.patch`.

One non-blocking note: the dashboard `api_setup()` creates the super agent with a local workspace, so it's not affected. But if the setup wizard ever gains workplace selection, the same helper should be called there.

Full review report saved at `artifacts/pr41-review.md`.

Best,
Richard
--
Robin Syihab's agent.

PR #42, opened 2026-05-16
test(zip): add regression tests for upload zip validation

DeryFerd:
## Bcakground

Plugin and skill uploads go through `backend/zip_validator.validate_upload_zip()` before extraction. The checks are non-trivial: compressed size limits, per-entry caps, path traversal, extension whitelist, and corrupt-archive handling. Until now that module had no dedicated unit tests, so a refactor or a well-meaning one-line change could weaken install-time defenses without anyone noticing in CI.

This PR adds a focused regression suite. It does not change production code.

## Cases covered

Each test builds small archives under a temp directory and calls `validate_upload_zip()` directly.

| Scenario | Expected |
|----------|----------|
| Typical skill layout (`skill.json` + `.py`) | Accept |
| `../` in entry name | Reject (traversal) |
| Absolute path (`/etc/passwd`) | Reject |
| Disallowed extension (`.exe`) | Reject |
| Empty archive | Reject |
| File that is not a zip | Reject |

Together these lock in the behaviors plugin/skill routes already rely on. They are cheap to run and fail with clear assertion messages on the error string returned by the validator.

## What I don't Fix

- **Cryptographic signing** of plugin packages — covered elsewhere ([#29](https://github.com/anvie/evonic/pull/29)).
- **Route-level** upload tests (multipart form, auth, disk paths) — could be a follow-up; this PR stays at the validator layer.
- Zip bomb limits that need hundreds of entries or huge uncompressed totals — not exercised here; existing constants in `zip_validator.py` are unchanged.

## How to verify

- `python -m pytest unit_tests/test_zip_validator.py -q`

### DIFF
--- a/unit_tests/test_zip_validator.py
+++ b/unit_tests/test_zip_validator.py
@@ -1,0 +1,68 @@
+"""Regression tests for backend.zip_validator."""
+
+import os
+import tempfile
+import unittest
+import zipfile
+
+from backend import zip_validator as zv
+
+
+class TestZipValidator(unittest.TestCase):
+    def setUp(self):
+        self.tmp = tempfile.mkdtemp()
+
+    def tearDown(self):
+        import shutil
+        shutil.rmtree(self.tmp, ignore_errors=True)
+
+    def _write_zip(self, name: str, entries: dict) -> str:
+        path = os.path.join(self.tmp, name)
+        with zipfile.ZipFile(path, 'w') as zf:
+            for arcname, content in entries.items():
+                zf.writestr(arcname, content)
+        return path
+
+    def test_accepts_valid_plugin_like_zip(self):
+        path = self._write_zip('ok.zip', {
+            'my-skill/skill.json': '{}',
+            'my-skill/tool.py': 'print("hi")',
+        })
+        ok, err = zv.validate_upload_zip(path)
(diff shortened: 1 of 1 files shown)


anvie (approved):
Thank you for the PR! This is a solid first-pass regression suite.

The six tests lock in the three most critical security checks: path traversal (`../`), absolute paths (`/etc/passwd`), and extension whitelisting (`.exe` blocked). The `_write_zip` helper keeps things clean, and `setUp`/`tearDown` with `tempfile.mkdtemp()` means no temp artifacts leak.

Good coverage for the happy path, empty zip, and non-zip file checks too. The assertions on error message substrings are pragmatic — more robust than exact matching.

Gaps worth noting (all acknowledged in your PR description):
- No corrupt-zip test (`BadZipFile`) — would be easy to add with a truncated file
- No Windows-style absolute path test (`\\evil\\file`) — the validator explicitly handles it
- Size-limit tests (50 MB upload, 500 entries, 200 MB uncompressed) are understandably out of scope

One minor nit: `import shutil` inside `tearDown` is functional but a bit unusual — moving it to module level would be cleaner.

Full review report saved at `artifacts/pr42-review.md`.

Best,
Richard
--
Robin Syihab's agent.

PR #44, opened 2026-05-17
fix(agents): setup super user agent failed due incorrect number of bindings supplied

ihfazhillah:
Background:
In the last step of initial setup, there is a 400 error. 

```
incorrect number of bindings supplied. the current statement uses 17, and there are 16 supplied.
```

The issue is due this change https://github.com/anvie/evonic/commit/7f6ce20e8edc6aba093c11ec242d1db004bb4b6f#diff-1abbadae03152c1ed03be415f07008f9948c4ab811cca9388a2aae15a25987e2

Changes:
`models/mixins/agents.py`


anvie:
Thank you!

PR #47, opened 2026-05-20
fix(tools): harden _skip_safety flag to require strict boolean True

DeryFerd:
Every tool backend (bash, write_file, str_replace, patch) checks `agent.get('_skip_safety')` to decide whether to run the heuristic safety pipeline. The issue: `.get()` returns truthy for almost anything. An LLM could set `_skip_safety` to the string `"true"`, the integer `1`, an empty dict `{}`, or even a list — and Python's truthiness would treat all of those as "skip safety."

This is a prompt injection vector. If an attacker can influence the agent context through a crafted prompt, they could convince the model to emit `_skip_safety: "true"` in its tool call arguments and bypass all safety checks — SSH path blocking, SQLite protection, sensitive system path guards, .env file protection, the whole pipeline.

The fix is a single helper function `should_skip_safety(agent)` in `backend/tools/lib/safety_pipeline.py` that uses an identity check: `agent.get("_skip_safety") is True`. Only the exact boolean `True` passes. Strings, integers, dicts, lists, empty values — all rejected. This means the flag can only be set by trusted server-side code (e.g. after explicit human approval), not by anything an LLM might generate.

I went through all four tool files and replaced every occurrence of the old pattern:

- `backend/tools/bash.py` — 1 occurrence
- `backend/tools/write_file.py` — 4 occurrences (SSH path, SQLite, sensitive paths, .env)
- `backend/tools/str_replace.py` — 3 occurrences (SSH path, sensitive paths, .env)
- `backend/tools/patch.py` — 3 occurrences (SSH path, sensitive paths, .env)

There's also a new test file with 11 cases covering the injection vectors: `None` agent, empty dict, explicit `True`, string `"true"`, string `"True"`, integer `1`, empty string, empty dict as value, empty list, boolean `False`, and a dict with other keys present. All pass.

**Backward compatibility:** If any legitimate code path was setting `_skip_safety` to a non-boolean truthy value, it will now stop skipping safety. That's the intended behavior — those paths should be reviewed and either set the proper boolean or not skip at all.

### DIFF
--- a/backend/tools/bash.py
+++ b/backend/tools/bash.py
@@ -57,9 +57,9 @@ def execute(agent: dict, args: dict) -> dict:
     # ------------------------------------------------------------------
     # HMADS safety check (pipeline: system rules + custom user rules)
     # ------------------------------------------------------------------
-    from backend.tools.lib.safety_pipeline import get_safety_pipeline
+    from backend.tools.lib.safety_pipeline import get_safety_pipeline, should_skip_safety
 
-    if not agent.get('_skip_safety') and agent.get('safety_checker_enabled', 1) and not agent.get('is_super'):
+    if not should_skip_safety(agent) and agent.get('safety_checker_enabled', 1) and not agent.get('is_super'):
         safety = get_safety_pipeline().check(script, tool_type='bash', agent_context=agent)
     else:
         safety = {'level': 'safe', 'score': 0, 'reasons': [], 'blocked_patterns': [], 'approval_info': {}}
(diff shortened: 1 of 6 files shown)


anvie (approved):
Nice catch. The `_skip_safety` flag with `is True` identity check is the right fix — truthy values pulling `None` through would've been a silent disaster. Clean, focused, well tested. This is how you do a security hardening patch.

Applied.

Best,
Linus T.
--
Robin Syihab's agent.

PR #48, opened 2026-05-20
fix(ssh): enable host key verification with known_hosts loading

DeryFerd:
The SSH backend in `backend/tools/lib/backends/ssh_backend.py` was using `paramiko.AutoAddPolicy()`, which silently accepts any host key presented by the remote server. No verification, no warning, no record. This makes every SSH workplace connection vulnerable to man-in-the-middle attacks — an attacker on the same network could intercept the connection, present their own host key, and capture all transmitted data including commands, file contents, and credentials.

Two changes:

1. **`AutoAddPolicy()` → `WarningPolicy()`**. Unknown hosts still connect (backward compatible — we don't want to break existing setups on first run), but paramiko now logs a warning instead of silently accepting. This at least makes the risk visible in logs.

2. **New `_load_host_keys()` method**. Before connecting, the backend now tries to load known host keys from standard locations: `~/.ssh/known_hosts` first, then `/etc/ssh/ssh_known_hosts` and `/etc/ssh/known_hosts` as fallbacks. If a host key is found in one of these files and the remote server presents a different key, paramiko will reject the connection rather than accepting it. If no known_hosts file exists anywhere, it falls back to `WarningPolicy` behavior and logs a hint suggesting the user run `ssh-keyscan -H <host> >> ~/.ssh/known_hosts`.

This mirrors how the `ssh` CLI works by default — it trusts keys it has seen before and warns about new ones. The difference is that before this change, the SSH backend was effectively running with `StrictHostKeyChecking no` on every connection.

**For users who already have SSH workplaces configured:** Their first connection after this upgrade will still succeed (via `WarningPolicy`), but they'll see a warning in the logs. If they want full verification, they can pre-populate known_hosts by connecting once via the `ssh` CLI or running `ssh-keyscan`.

**For new connections:** Same behavior as before, just with a visible warning instead of silent acceptance.

Three tests verify that `WarningPolicy` is used, `AutoAddPolicy` is gone, and the `_load_host_keys` method exists on `SSHBackend`.

### DIFF
--- a/backend/tools/lib/backends/ssh_backend.py
+++ b/backend/tools/lib/backends/ssh_backend.py
@@ -54,8 +54,43 @@ class SSHBackend(ExecutionBackend):
         self._evonic_installed = False
 
         self._client = paramiko.SSHClient()
-        self._client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        self._load_host_keys()
+        self._client.set_missing_host_key_policy(paramiko.WarningPolicy())
         self._connect()
+
+    def _load_host_keys(self):
+        """Load known_hosts from standard locations for host key verification.
+
+        Tries the user's ~/.ssh/known_hosts first, then falls back to the
+        system-wide known_hosts. If neither exists, the client will still
+        connect (via WarningPolicy) but will log a warning for unknown hosts.
+        """
+        import paramiko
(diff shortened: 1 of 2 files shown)


DeryFerd:

PR #51, opened 2026-05-29
fix(kanban): normalize task_id to string to prevent type mismatch bug

Rainfalldrop:
## Summary

This PR fixes the kanban plugin bug where `state('kanban:activate')` returns "user has not approved yet" error even when the user has explicitly approved.

## Root Cause

**Type mismatch between integer (database) and string (agent function params) task IDs:**

- Database returns task IDs as integers: `123`
- Agent tool calls pass task IDs as strings: `"123"`
- In-memory dictionary comparisons like `_approval_granted.get(agent_id) != task_id` fail because `123 != "123"`
- This causes the approval check to always fail, resulting in "user has not approved yet" error

## Changes

1. **Normalize task_id to string at extraction** in `_state_handler()`:
   ```python
   # Before
   task_id = data.get('task_id') if isinstance(data, dict) else data
   # After
   task_id = str(data.get('task_id')) if isinstance(data, dict) else str(data)
   ```

2. **Normalize task_id to string** when storing to `_pending_tasks` and `_active_tasks` dictionaries (all 5 locations)

3. **Fix comparison logic** to use `str()` wrapper when checking `_approval_granted`:
   ```python
   # Before
   if _approval_granted.get(agent_id) != task_id:
   # After
   if str(_approval_granted.get(agent_id)) != task_id:
   ```

## Files Changed

- `plugins/kanban/handler.py` (14 insertions, 14 deletions)

## Testing

A test script was created to verify:
- String-to-string comparison works correctly (approved ✓)
- Old integer-to-string comparison would fail (bug ✓)
- The fix resolves the mismatch

## Related Issue

The bug manifests when:
1. Agent acknowledges a task (`kanban:pick`)
2. Agent presents task to user
3. User approves ("ya/boleh")
4. Agent calls `kanban:activate` → **FAILS with "user has not approved yet"**
5. On second attempt, it may succeed (approval was set by message interceptor from previous turn)

This fix ensures approval is properly recognized on the first `activate` call by normalizing task_id to a consistent string type throughout.

### DIFF
--- a/plugins/kanban/handler.py
+++ b/plugins/kanban/handler.py
@@ -1135,7 +1135,8 @@ command_registry.register(
 def _state_handler(agent_id: str, session_id: str, agent_state, label: str, data):
     """Handle kanban workflow state transitions via the state() built-in tool."""
 
-    task_id = str(data.get('task_id')) if isinstance(data, dict) else str(data)
+    raw = data.get('task_id') if isinstance(data, dict) else data
+    task_id = str(raw) if raw is not None else None
 
     # ── kanban:pick ───────────────────────────────────────────────────────────
     if label == 'kanban:pick':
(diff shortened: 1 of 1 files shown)


anvie:
Ini banyak tests yang failed, bisa tolong diperbaiki dulu?

PR #55, opened 2026-06-03
fix(install.sh): shallow clone breaks git fetch/pull — reconfigure remote to track branches

binsarjr:
### Bug: Shallow clone locks remote to single tag forever

`install.sh` clones using:

```bash
git clone --depth 1 --branch "$LATEST_TAG" ...
```

This produces a remote config locked to a single tag:

```
fetch = +refs/tags/v0.5.24:refs/tags/v0.5.24
```

As a result, **`git fetch`, `git pull`, and the supervisor auto-updater are permanently broken** because no branch refs are tracked.

### Evidence that this is a bug (not by design)

1. **After cloning, the script intentionally checks out `main`**: [install.sh#L137-L138](https://github.com/anvie/evonic/blob/main/install.sh#L137-L138)
   ```bash
   # Ensure we're on the main branch so users can git pull manually
   git -C "$EVONIC_HOME" checkout main 2>/dev/null
   ```
   The developer **intended** for users to be able to `git pull` — but the remote config produced by the shallow clone prevents it.

2. **The update path calls `git fetch` and `git pull`**: [install.sh#L108-L116](https://github.com/anvie/evonic/blob/main/install.sh#L108-L116) — these commands are assumed to work but silently fail because the remote only tracks a single tag.

3. **Evonic's supervisor auto-updater** depends on `git fetch` — the auto-update feature is completely broken.

### Fix

The `fix_remote_fetch()` function detects remotes locked to a single tag (`refs/tags/*:refs/tags/*`) and repairs them:

```bash
git config remote.origin.fetch "+refs/heads/*:refs/remotes/origin/*"
git config --add remote.origin.fetch "+refs/tags/*:refs/tags/*"
git fetch origin
```

Called in the **update path** (fix existing broken installs) and **after every clone path** (new installs).

### Impact

After the fix, users can:
- `git pull` for manual updates
- `git fetch` to see new updates
- The supervisor auto-updater works normally


anvie:
Closing in favor of #59 — the original branch carried unrelated commits (caused by agent rebase incidents on main branch), so I made a fix using git push force. A clean replacement has been opened with only the relevant change.

PR #58, opened 2026-06-04
feat(health): add DB, disk, and Docker checks to health endpoint

DeryFerd:
### Problem

The `/api/health` endpoint has always returned three fields: `status`, `uptime`, and `version`. That is enough to know the process is running, but not enough to know whether it is actually working. The update supervisor calls this endpoint after staging a release to decide whether to swap the symlink. If the database is down or the disk is full, the process still returns `{"status": "ok"}` and the supervisor proceeds like nothing happened.

### What changed

The endpoint now runs three checks on every request and includes a `checks` object alongside the existing fields:

```json
{
  "status": "ok",
  "uptime": 1234.5,
  "version": "1.2.3",
  "checks": {
    "database": "ok",
    "disk": {
      "total_gb": 100.0,
      "free_gb": 45.2,
      "free_pct": 45.2
    },
    "docker": "24.0.7"
  }
}
```

If any check fails, `status` changes from `"ok"` to `"degraded"`. This lets the supervisor (or any monitoring tool) distinguish between "the process is running" and "the process is running and its dependencies are healthy."

### The database check

It runs a minimal query — `db.get_agents()` — and catches any exception. SQLite can go readonly, the file can get deleted, or the connection can lock up. Any of those return `"database": "error"` and flip the overall status to degraded.

### The disk check

`shutil.disk_usage()` on the app root directory. Returns total, free, and free percentage. If the call fails (unlikely on normal systems, but possible in locked-down containers where `os.stat` is restricted), it returns an empty object rather than crashing the whole endpoint.

### The Docker check

This one was harder to decide on. Running `docker info` on every health poll is not free, but five seconds with a timeout felt reasonable given that this endpoint is called once or twice during an update cycle, not every few seconds under load. If Docker is not installed or the socket is unreachable, it returns `"unavailable"`. If the subprocess itself errors out (OOM, signal, etc), it returns `"unknown"`.

I considered making the Docker check opt-in behind an env var, but keeping it always-on means the supervisor gets a clear signal when a staged release that depends on sandbox containers cannot actually reach the daemon. If latency becomes a problem later, it is easy to swap the implementation for a socket check or cache the result.

### Backward compatibility

The old `status`, `uptime`, and `version` fields are still present with the same types. The new `checks` field is additive. Any existing monitor that only looks at `status` will still see `"ok"` — unless a check fails, in which case `status` now says `"degraded"` instead. That is the one behavioral difference. If someone has an alert that literally checks `status == "ok"`, it still fires. If someone checks `status != "ok"`, it now catches dependency failures too.

The seven existing unit tests in `test_health.py` all pass without changes.


anvie:
Closing in favor of #61 — the original branch carried unrelated commits (caused by agent rebase incidents on main branch), so I made a fix using git push force. A clean replacement has been opened with only the relevant change.

PR #67, opened 2026-06-09
fix(auth): make ProxyFix trust count configurable to prevent IP spoofing

DeryFerd:
## Problem

`app.py` hardcodes `ProxyFix(x_for=1)`, which tells Flask to always trust the rightmost entry in the `X-Forwarded-For` header. That works when there is exactly one reverse proxy in front of Evonic (nginx, Caddy, Cloudflare, etc.), which is the common case.

But it breaks for direct deployments. If someone exposes Evonic straight to the internet without a reverse proxy, any client can send a fake `X-Forwarded-For` header and Evonic will trust it. The real client IP becomes whatever the attacker chose. This defeats the login rate limiter in `routes/auth.py` (keyed on `request.remote_addr`) and pollutes the security audit logs in `backend/security_audit.py` with wrong IPs.

## The fix

Adds a `PROXY_TRUST_COUNT` environment variable so operators can match the trust level to their deployment topology:

- `0` for direct deployments with no proxy. `X-Forwarded-For` is ignored entirely.
- `1` (default) for single proxy setups, same behavior as today.
- `2` or higher for multi proxy chains (CDN in front of nginx in front of the app).

## What changed

- `config.py` defines `PROXY_TRUST_COUNT` with bounds checking (0 to 10, default 1) via `_get_env_int`.
- `app.py` reads `config.PROXY_TRUST_COUNT` instead of the hardcoded `1`.
- `unit_tests/test_proxy_trust.py` covers config parsing edge cases (default, zero, two, negative clamped to zero, invalid falls back to default) plus actual IP resolution behavior for trust=0 and trust=1.

## Testing

All 7 new tests pass. Existing config and health tests still pass.

## Backward compatibility

Default stays at 1, so existing deployments behind a proxy don't need to change anything. If you run Evonic directly on the internet, add `PROXY_TRUST_COUNT=0` to your `.env`.

### DIFF
--- a/app.py
+++ b/app.py
@@ -70,9 +70,19 @@ app.config['PERMANENT_SESSION_LIFETIME'] = timedelta(days=7)
 # Global upload size limit (defense-in-depth for all endpoints)
 app.config['MAX_CONTENT_LENGTH'] = 50 * 1024 * 1024  # 50 MB
 
-# Trust proxy headers from Cloudflare / nginx / any reverse proxy
+# Trust proxy headers based on deployment topology.
+# PROXY_TRUST_COUNT=0 for direct deployments (no reverse proxy) to prevent
+# IP spoofing via X-Forwarded-For. Set to 1 for single-proxy (default), or
+# higher for multi-proxy setups (CDN → reverse proxy → app).
+# See config.py for full documentation.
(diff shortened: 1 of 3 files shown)


DeryFerd:

PR #70, opened 2026-06-18
fix: eval engine bugs — tool_calling errors, reasoning extractors, In…

txtsamu:
…donesian number parsing

- engine.py: redirect pass2 answer_extractor client to the run's LLM so it doesn't fall back to the offline default at port 8000; also add missing system_prompt=None before tool_calling branch to fix NameError
- tools.py: add write_file tool definition and mock handler so tool_calling L4 (and any test that calls it) doesn't fail with unknown tool
- strategies/tool_call.py: recognise "tools_used" key in legacy test expected dicts (alongside existing "tool", "tools", "chain")
- answer_extractor.py: fix reasoning L2 extractor (boolean→sequence, number-sort prompt) and L4 extractor (boolean causality→statements, multiple-choice prompt)
- test_classes/reasoning.py: fix Indonesian thousands-separator parsing in L5 scorer — single dot with exactly 3 trailing digits (e.g. "820.800") was not being stripped, causing 820.8 instead of 820800

### DIFF
--- a/evaluator/answer_extractor.py
+++ b/evaluator/answer_extractor.py
@@ -49,20 +49,21 @@ Your answer (ya/tidak only):""",
         },
         
         2: {
-            "template": """You are given a question and an AI's response. Extract ONLY the final answer: "ya" or "tidak".
+            "template": """You are given a question asking to sort numbers, and an AI's response. Extract ONLY the sorted numbers as a comma-separated list.
 
 ---BEGIN RESPONSE---
 {response}
 ---END RESPONSE---
 
 Rules:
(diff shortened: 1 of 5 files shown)


anvie:
Can you please fix the failing tests?

txtsamu:
The test failure is a pre-existing Python 3.9 incompatibility in `backend/tools/kb_graph.py` that was fixed in upstream main (commit f6548ca7, Jun 19) — unrelated to this PR's changes. I've pushed a retrigger commit; just need maintainer approval to run workflows. The CI should pass with the updated merge base.

PR #72, opened 2026-06-20
fix: resolve fresh-install issues (dashboard, doctor, installer)

srflmr:
## Summary
On a fresh v0.8.0 install, the dashboard landing page fails to load and the installer can abort with a cryptic error on common setups. This PR fixes the broken queries, adds a diagnostic to catch the regression class, and hardens the installer.

## Dashboard
- **`get_model_usage` 500 on fresh install.** The query selected `COALESCE(model, …)`, but the agents `model` column was removed in v0.6.78 (renamed to `model_id`). Existing databases keep the old column, so the bug was masked; fresh installs never create `model`, so the query raised `no such column: model` and 500'd both `/api/dashboard/data` and `/api/dashboard/stats` — the landing page. Point the query at `model_id`; the `AS model` alias preserves the response shape (no API change).
- **Recent-agents model badge** read `a.model` (always `undefined` after the rename) → use `a.model_id`.
- **Empty states.** The dashboard widgets reused their skeleton loaders as the "no data" state, so on a fresh install (no agents/models/evaluation runs) they animated indefinitely. Show a "no data yet" message instead.

## Doctor
- Add a **Database Schema / Query Health Check** that runs the dashboard read queries against the live schema. A plain `SELECT 1` connection check can't catch a column rename/drop that leaves a query stale — this does. Runs read-only in both normal and `--fix` mode.

## Installer (`install.sh`)
- **Verify the `venv`/`ensurepip` module** (the real requirement) instead of an unused system `pip3`, failing fast with guidance to install `python3-venv`. On distros that create a venv without bootstrapping pip, `python3 -m venv` returns 0 yet leaves no pip, so dependency install previously failed later with a cryptic `pip3: not found`.
- **Bootstrap pip via `ensurepip`** if the venv comes up without it; install dependencies via `python -m pip` rather than a `pip`/`pip3` shim.
- **Activate `.venv` first, then legacy `venv`** in the generated wrapper, matching the repo `evonic` wrapper (`.venv` primary, `venv` fallback).
- Drop non-POSIX `local` (the script is `#!/bin/sh`); verified with `sh -n`, `dash -n`, and `bash -n` (the `curl … | bash` one-liner).

## Testing
- Reproduced the dashboard 500 on a fresh DB; confirmed `/api/dashboard/data` and `/api/dashboard/stats` return 200 after the fix.
- Added `unit_tests/test_dashboard_model_usage.py` (fails on the old query, passes on the fix). Full suite: no new failures.
- `evonic doctor` and `evonic doctor --fix` run the new schema check cleanly.

## Scope
5 files — query/UI/installer fixes plus one regression test. No schema migration, no API changes.

### DIFF
--- a/cli/commands.py
+++ b/cli/commands.py
@@ -3179,6 +3179,37 @@ def doctor_command(quick=False, fix=False, with_llm_provider=False):
         results.append(_fail(f"Asset build check failed: {e}"))
 
 
+    # ── 13. Database Schema / Query Health Check ──────────────
+    _section("13. Database Schema / Query Health Check")
+    try:
+        import sqlite3 as _sqlite3
+        from models.db import db as _schema_db
+
+        dashboard_queries = [
+            ("dashboard stats", _schema_db.get_dashboard_stats),
+            ("model usage", _schema_db.get_model_usage),
+            ("model leaderboard", _schema_db.get_model_leaderboard),
+            ("recent agents", _schema_db.get_recent_agents),
+            ("recent runs", _schema_db.get_recent_runs),
+        ]
+        schema_ok = True
+        for q_name, q_fn in dashboard_queries:
+            try:
+                q_fn()
+            except _sqlite3.OperationalError as qe:
+                schema_ok = False
+                results.append(_fail(f"{q_name} query failed — schema mismatch: {qe}"))
+            except Exception as qe:
+                results.append(_warn(f"{q_name} query raised {type(qe).__name__}: {qe}"))
+        if schema_ok:
+            results.append(_ok("Dashboard queries match the current database schema"))
+        else:
+            _info("  A column referenced by a query is missing from the schema "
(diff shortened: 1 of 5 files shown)


srflmr:
Related: #73 — while investigating fresh-install behavior I also found that the `evomem` binary isn't shipped, so the default memory engine silently falls back to FTS5. That's a packaging concern (only maintainers can ship the binary), so I filed it separately rather than folding it into this PR.

anvie:
Thanks for your contribution.

Owh, the evomem bin..., I almost forgot, I’ve published the Evomem repo at https://github.com/anvie/evomem
The pre-compiled binary for Linux and Darwin are also available there, feel free to use

srflmr:
Thanks! I'll use the prebuilt binary from [anvie/evomem v0.2.0](https://github.com/anvie/evomem/releases/tag/v0.2.0) — that resolves #73 on my end.

PR #75, opened 2026-06-22
fix: in-place restart, health-gated update/rollback, Python 3.14 test compat

srflmr:
> 📋 **Sequential PR series — 1 of 2**
> 1. **fix: restart/update reliability ← THIS PR** (base: `main`)
> 2. feat: evomem provisioning + Discord _(builds on this; opened after this merges)_
>
> A follow-up feature PR depends on this one (its dashboard auto-restart uses `backend/restart.py` introduced here). I'll open it after this merges so its diff stays clean. Happy to split/reorder per your preference.

---

Three correctness/reliability fixes for the self-update + restart path. Commits are kept separate for review.

### 1) `fix(restart)`: single in-place re-exec helper
**Problem.** The web Update "Restart" button (`update_manager.trigger_restart`) sent `SIGTERM` to the parent and assumed an external process manager (systemd/Docker) would respawn it. The documented install path is `evonic start -d`, which spawns `python app.py` detached with **no** supervisor; `AgentRuntime._signal_handler` turns SIGTERM into `sys.exit(0)`, so the server stayed **down** with a stale PID file. Three other restart paths (`/restart`, super-agent restart tool, web setup auto-restart) already re-exec'd in place, but that logic was copy-pasted with subtle drift — which is how the broken fourth copy went unnoticed. None of them shut the scheduler down.

**Change.** New `backend/restart.py` as the single source of truth (`restart_in_place` / `schedule_restart`): stop channels (so Telegram releases its long-poll before re-bind), shut down the scheduler, free inherited FDs, then `os.execv` against `app.py` — preserving the PID the lock/PID file point to, so it works with or without a process manager. All four call sites repointed, each keeping only its own pre-restart bookkeeping.

### 2) `fix(update)`: health-gated update + deterministic rollback
**Problem.** The web updater ran `git fetch` + `git reset --hard` only — no dependency reinstall (an upstream requirement bump broke the next boot) and no check that the new tree imports. The CLI did git + `doctor --fix` but also no deps reinstall, and hard-failed if doctor returned non-zero. Rollback reset to the reflog's `HEAD@{1}`, which a failed-update auto-rollback corrupts, and `evonic update --rollback` was a dead flag.

**Change.** One shared, health-gated pipeline used by both web and CLI: `reset --hard <ref>` → reinstall requirements → `doctor --fix` → import smoke test. deps and the smoke test are gating; on failure the tree resets back to the recorded prior commit (auto-rollback), so a broken update never takes the running server down. `doctor --fix` is best-effort (logged, non-gating) — the smoke test is the real boot gate. The smoke test runs `import app` in a subprocess with `EVONIC_SMOKE_TEST=1` (skips channel/scheduler startup; `app.run()` is under `__main__`), so it has no side effects on the live server. Rollback is deterministic (recorded `previous_commit`; reflog kept as fallback) and `--rollback` is wired. The two-step UX is preserved (update applies code; the admin restarts via the now-working action); super-agent/supervisor flows are unchanged.

### 3) `fix(tests)`: Python 3.14 compatibility for `agent_messaging`
**Problem.** `setup_module` replaced `models` / `backend.agent_runtime` with bare `MagicMock`s; Python 3.14's import machinery reads `parent.__spec__` while resolving `from models.X import ...`, and MagicMock raises `AttributeError` for unset dunders — 61 errors at setup.

**Change.** Use `types.ModuleType` stand-ins (`__spec__=None`) and attach the behavioural mocks as attributes; assertions unchanged. Test-only.

### Why it's safe
- Restart **decision logic is untouched** — inter-agent approval guard, `restart_ready_needed` / `restart_greeting_needed`, and context capture; only the `_do_restart` mechanics moved into the shared helper. `[python, app.py]` equals the daemon's real argv; `scheduler.shutdown()` is additive (guarded).
- Update keeps the deliberate two-step UX; no auto-restart introduced.
- The test change is portable (no 3.14-only idiom) and raises nothing about the minimum Python.

### Test plan
- Full unit suite: **1093 passed / 0 failed** on **Python 3.10 and 3.14** (this fixes `main`'s 61 errors).
- Manual: web "Restart Now" under `evonic start -d` returns on the same PID/port; restart twice with Telegram enabled → no `Conflict`; simulated broken update → auto-rollback, live server stays up; `evonic update --rollback` restores the prior commit.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -52,3 +52,9 @@ EVONIC_SUBAGENT_IDLE_TTL=1800
 
 # Session Archive: when enabled the session data will be archived before it deleted in active db
 EVONIC_SESSION_ARCHIVE=0
+
+# Boot-gate probe: set to 1 to make `import app` return immediately without starting
+# channels, the scheduler, or app.run(). Used by the update pipeline's smoke test to
+# verify the new code imports cleanly without touching the live server's port or DB
+# connections. Not intended for production use.
+# EVONIC_SMOKE_TEST=
--- a/app.py
+++ b/app.py
@@ -319,7 +319,10 @@ init_super_agent_notifier()
 import os as _os
 _reloader_active = _os.environ.get('WERKZEUG_RUN_MAIN') is not None
 _is_reloader_child = _os.environ.get('WERKZEUG_RUN_MAIN') == 'true'
-if not _reloader_active or _is_reloader_child:
+# Smoke-test import (used by the updater to validate a new tree before
+# restarting): import all modules but skip channel/scheduler startup.
+_smoke_test = _os.environ.get('EVONIC_SMOKE_TEST') == '1'
+if (not _reloader_active or _is_reloader_child) and not _smoke_test:
     # Run SYSTEM.md migration eagerly (not lazily on first GET /api/agents).
     # Agents that predate the on-disk SYSTEM.md feature need their file written
     # before they start processing messages — otherwise read_file("/_self/SYSTEM.md")
--- a/backend/restart.py
+++ b/backend/restart.py
@@ -1,0 +1,80 @@
+"""
+In-place server restart — single source of truth.
+
+Restarts the running server by replacing the current process image with a fresh
+one (os.execv), preserving the PID the run/ lock and PID file point to. This is
+the only restart mechanism that works without an external process manager, so it
+is correct under the documented `evonic start -d` install path as well as under
+systemd/Docker.
+
+Used by the /restart slash command, the super-agent restart tool, the web setup
+auto-restart, and the web update "Restart" action so every restart path behaves
+identically.
+"""
+
+import logging
+import os
+import sys
+import threading
+import time
+
+log = logging.getLogger(__name__)
+
+
+def restart_in_place(delay: float = 1.5) -> None:
+    """Stop channels and the scheduler, free inherited FDs, and re-exec the server.
+
+    Runs in the calling thread and never returns on success (the process image is
+    replaced). Intended to be invoked from a daemon thread via schedule_restart().
+    """
+    time.sleep(delay)  # Let the triggering response/log flush first.
+
+    # Stop all channels cleanly so Telegram releases its long-poll before the
+    # new process re-opens them (otherwise the bot token hits a getUpdates
+    # Conflict on the next boot).
+    try:
+        from backend.channels.registry import channel_manager
+        channel_manager.stop_all()
+        time.sleep(1.0)  # Give Telegram server-side time to release.
+    except Exception as e:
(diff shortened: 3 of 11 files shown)


srflmr:
Hi @anvie — thanks for maintaining evonic! These two PRs fix a few things I ran into while using the project. Happy to address any feedback or split further if you'd prefer a different shape.

When you get a chance, could you review **#75 first** (restart/update/rollback reliability), then **#76** (evomem provisioning + Discord)? #76 builds on #75's `backend/restart.py`, so the sequential order matters. I'll rebase #76 onto main after #75 merges to clean up its diff.

No rush — appreciate the work you've put into this project.

anvie:
Thank you for this PR! The consolidation of restart logic into a single
`restart_in_place()` entry point is excellent!

A few things to address before merging:

- Rebase needed - there’s a merge conflict with `main`. Please rebase
  so the diff can be reviewed cleanly.
- Concurrent update safety - `apply_update()` has no locking. A
  concurrent call from the web UI and CLI could race on git operations.
  A lightweight file lock or pidfile guard would be enough for now.
- WebNotifier granularity - UI progress went from 2 coarse steps to…
  2 coarse steps. Exposing the reinstall and smoke-test phases through the
  notifier would give operators better visibility during long updates.
- `EVONIC_SMOKE_TEST` env var - worth adding an example in `.env.example` so user know
  it exists.

Overall, this is a well-structured refactor with the right test coverage.
After rebase and the lock guard, it’s good to merge.

srflmr:
Thanks for the review @anvie! Addressed all 4 items:

**✅ Rebased onto current main**
Conflict in `trigger_restart()` resolved — incorporated your fix from a8dc3e7 (state reset to idle before re-exec) while replacing the detached subprocess with the unified `schedule_restart()` helper from `backend/restart.py`.

**✅ Concurrent update safety**
Added `_acquire_update_lock()` / `_release_update_lock()` using `fcntl.flock(LOCK_EX | LOCK_NB)` on POSIX (no-op on Windows). Lock file lives in `state/update/update.lock` alongside `update_state.json`. Both `apply_update()` and `apply_rollback()` hold the lock for the duration of git operations, so a web UI trigger and a CLI call can't race on the working tree.

**✅ WebNotifier granularity**
`TOTAL_STEPS` is now 5 and `apply_update()` accepts an optional `progress_cb(step, total, label)` parameter. The web thread passes `notifier.send_progress` as the callback; the CLI path calls `apply_update(target)` unchanged. Steps reported: fetch → reset → reinstall deps → doctor --fix → smoke test.

**✅ `EVONIC_SMOKE_TEST` in `.env.example`**
Documented with a comment explaining the boot-gate mechanic and that it's not for production use.

PR #76, opened 2026-06-22
feat: evomem auto-provisioning + setup wizard + Discord channel

srflmr:
> 📋 **Sequential PR series — 2 of 2** · **depends on #75**
> 1. #75 — fix: restart/update reliability _(merge first)_
> 2. **feat: evomem provisioning + Discord ← THIS PR**
>
> ⚠️ **Stacked on #75.** Until #75 is merged, the diff below also includes #75's
> 3 fix commits. The **3 commits that belong to this PR** are:
> - `ad1727d` feat(evomem): auto-provision the memory-engine binary
> - `02b7386` feat(setup): add evonic setup CLI wizard with evomem provisioning
> - `76566c2` feat: add Discord channel support
>
> This PR's dashboard auto-restart reuses `backend/restart.py` from #75. Once #75
> merges I'll rebase onto the updated `main` and the diff will narrow to just the
> 3 feat commits. Happy to hold review until then.

---

Two additive features (commits kept separate for review).

### 1) `feat(evomem)`: auto-provision the memory-engine binary
**Problem.** The memory engine defaults to evomem, but nothing installed the binary, so fresh installs silently fell back to FTS5; installer, doctor, and setup were inconsistent about it.

**Change.** `backend/evomem_provision.py` as the single source of truth: resolve the latest release via the GitHub API (`EVOMEM_VERSION` pins a tag), verify the download against the API-reported SHA-256, install atomically. On any failure (unsupported platform, network, checksum) it returns cleanly and the runtime keeps using FTS5. Wired through every surface: CLI `evonic evomem install [--force]`; `doctor --fix` (installs before judging the memory-engine check, instead of only creating the directory); `install.sh` (best-effort, non-fatal); `setup` (logs whether evomem is ready or FTS5 will be used); `.env.example` documents `EVONIC_MEMORY_ENGINE` / `EVOMEM_VERSION` / `EVOMEM_BINARY`.

### 2) `feat(setup)`: `evonic setup` CLI wizard
**Problem.** `install.sh` called `evonic setup`, but no such subcommand existed — under `set -e` the installer exited non-zero, leaving fresh installs unconfigured.

[...]

PR #78, opened 2026-06-23
feat: evomem memory-engine auto-provisioning + setup integration

srflmr:
Auto-provision the evomem memory-engine binary and wire it into the setup surfaces. Two commits, kept separate for review.

> Supersedes #76 — that PR bundled a Discord channel that shares no code with this change. Discord is now its own PR; this one is evomem + setup only.

### 1) `feat(evomem)`: auto-provision the memory-engine binary
**Problem.** The memory engine defaults to evomem, but nothing installed the binary, so fresh installs silently fell back to FTS5; installer, doctor, and setup were inconsistent about it.

**Change.** `backend/evomem_provision.py` as the single source of truth: resolve the latest release via the GitHub API (`EVOMEM_VERSION` pins a tag), verify the download against the API-reported SHA-256, install atomically. On any failure (unsupported platform, network, checksum) it returns cleanly and the runtime keeps using FTS5. Wired through every surface: CLI `evonic evomem install [--force]`; `doctor --fix` (installs before judging the memory-engine check, via the single `ensure_evomem()` — no redundant directory stub); `install.sh` (best-effort, non-fatal); `run_setup()` logs whether evomem is ready or FTS5 will be used; `.env.example` documents `EVONIC_MEMORY_ENGINE` / `EVOMEM_VERSION` / `EVOMEM_BINARY`.

### 2) `feat(setup)`: provision evomem during setup
**Change.** Hook provisioning into the existing setup surfaces — does **not** add a parallel wizard; builds on your `evonic setup` from `b3bdd37`:
- CLI `evonic setup`: an evomem install step (interactive, default yes) after the super agent is created. Skipped under `--non-interactive`; `install.sh` provisions evomem directly and `ensure_evomem()` is idempotent.
- Web setup wizard (`templates/setup.html`): an "Install evomem" checkbox (default on) + a Review row; `routes/dashboard.py` reads `install_evomem` and runs `ensure_evomem()` in the background restart thread so the download never blocks the HTTP response.

### Why it's safe
- Additive — new module + an opt-in step on existing commands. No `models/` schema changes.
- evomem provisioning is **opt-in** (default yes) and **best-effort** with a clean FTS5 fallback; no surface fails when the binary can't be installed.
- Rebased on current `main`; reuses your `setup_command()` and `backend/restart.py`.

### Test plan
- `test_evomem_provision` green (network mocked).
- `python -m backend.evomem_provision` downloads + installs evomem (best-effort, FTS5 fallback otherwise).
- `evonic setup`, `evonic evomem install`, `doctor --fix` exercised cleanly. `evonic setup --help` confirms no subparser collision with the upstream wizard.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -58,3 +58,13 @@ EVONIC_SESSION_ARCHIVE=0
 # verify the new code imports cleanly without touching the live server's port or DB
 # connections. Not intended for production use.
 # EVONIC_SMOKE_TEST=
+
+# Memory engine: "evomem" (default, graph-backed; requires the evomem binary) or
+# "fts5" (SQLite full-text fallback, no binary needed). The installer fetches the
+# evomem binary automatically; otherwise run `evonic evomem install`. When evomem
+# is selected but the binary is absent, the runtime transparently uses FTS5.
+EVONIC_MEMORY_ENGINE=evomem
+# Pin a specific evomem release tag (default: latest). e.g. EVOMEM_VERSION=v0.2.0
+# EVOMEM_VERSION=
+# Override the evomem binary path (default: shared/bin/evomem).
+# EVOMEM_BINARY=
--- a/backend/evomem_provision.py
+++ b/backend/evomem_provision.py
@@ -55,8 +55,14 @@ _DOWNLOAD_TIMEOUT = 60
 
 
 def default_binary_path() -> str:
-    """Canonical install path for the evomem binary: <repo>/shared/bin/evomem."""
-    return os.path.join(_BASE_DIR, "shared", "bin", _BINARY_NAME)
+    """Install path for the evomem binary.
+
+    Honours the EVOMEM_BINARY override so the provisioner writes exactly where
+    evomem_client._resolve_binary() reads it; otherwise <repo>/shared/bin/evomem.
+    Without this, setting EVOMEM_BINARY would install the binary to a path the
+    runtime never consults.
+    """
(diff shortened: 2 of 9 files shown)


anvie:
Thanks for the PR! The single source of truth pattern for evomem binary provisioning is a great design choice, ensuring consistency across the CLI, web setup wizard, and installer. The implementation correctly handles idempotency, atomic installation, and integrity verification with thorough test coverage. One minor suggestion: consider downgrading the logging severity in `_fail()` from `ERROR` to `WARNING` for expected conditions like unsupported platforms to reduce production log noise. Overall, this looks great and is ready to merge.

srflmr:
Thanks for the review, @anvie!

Addressed the logging-severity suggestion: `_fail()` now takes a `level` arg (default `ERROR`). The expected, non-actionable conditions — unsupported platform and "no release asset for this platform" — log at `WARNING` to cut production noise, while genuine failures (network errors, missing sha256 digest, checksum mismatch, install error) stay at `ERROR` so real problems remain visible.

While in there I rebased onto the latest `main` and added one small related fix: `default_binary_path()` now honours the `EVOMEM_BINARY` override, mirroring `evomem_client._resolve_binary()`. Previously it always installed to `shared/bin/evomem`, so a custom `EVOMEM_BINARY` would be provisioned to a path the runtime never reads — now the provisioner writes exactly where the engine looks.

Tests assert both log severities and the env-override path; full suite green. Pushed as a separate commit on top so the changes are easy to diff.

PR #84, opened 2026-06-26
fix: handle Qwen models putting response in reasoning_content

ihfazhillah:
When Qwen models run via llama.cpp without --reasoning, they put the entire response in reasoning_content with an empty content field.

This adds a fallback in both llm_loop.py and llm_client.py to treat reasoning_text as the actual response content when:
- content is empty
- no embedded_final from thinking tags
- no tool calls were generated

This fixes silent failures where agents using certain Qwen models appear to not respond.


anvie:
PR ini jadi regression ternyata:

ss_004346

PR #88, opened 2026-07-15
feat: improve ChatView command and Bash modes

binsarjr:
## Summary
- Add a dynamically discovered, keyboard-accessible slash-command palette to the ChatView composer.
- Position command suggestions above the composer and insert a selected command without submitting it.
- Refine `!` Bash mode as a red, monospace shell state with a compact helper message above the composer.
- Preserve the existing server-side Bash permission and web-session enforcement.

## Validation
- Jinja template compilation
- Extracted ChatView composer JavaScript syntax check with `node --check`
- `git diff --check`
- Command API smoke tests, including unknown-agent `404`

## Notes
- The full pytest suite was not run because pytest was unavailable in the local system and virtual environments.

### DIFF
--- a/backend/slash_commands.py
+++ b/backend/slash_commands.py
@@ -44,6 +44,33 @@ class SlashCommandRegistry:
         return [(cmd.name, cmd.description) for cmd in self._commands.values()]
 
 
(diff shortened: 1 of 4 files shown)


binsarjr:
image
image

PR #98, opened 2026-08-05
feat(agentapi): stream coalesced tool-call events in SSE chat responses

fepfitra:
## Problem

The agentapi streaming chat endpoint (`/plugin/agentapi/v1/chat/completions` with `stream: true`) only forwards `llm_response_chunk` and `turn_complete` events. Streaming clients (e.g. the pitura-dashboard) can show the agent's text but have NO visibility into what the agent is DOING (running bash, editing files, searching the web). Tool execution is a black box during long agent turns.

## Fix

Subscribe to the existing `tool_executed` event stream, coalesce tool calls per session into a 5-second window (mirroring the Telegram channel pattern from PR #95), and emit inline SSE events while the stream is open:

```json
data: {"type":"tool","tool_name":"bash (2x), python (1x)","has_error":false}
```

- A 100-tool loop collapses to a handful of lines, not one per call.
- `has_error` surfaces failures (client can style them ⚠️).
- The `type: "tool"` field distinguishes tool events from normal `choices[].delta.content` chunks.
- Cleanup: `tool_executed` is unsubscribed in the same `finally` block as the other events.

## Verification

`python -c 'import ast; ast.parse(...)'` — syntax OK. Live-tested against the dashboard after restart: tool calls appear as compact chips in the chat while the agent works.


fepfitra:
Closing: PR was polluted with 12 unrelated commits from merge/our-fixes (clone-workplace, cache-invalidation, telegram-notify, llm_client UA, skill-allowlist). Reopening as a clean single-commit PR.

Issue #45, opened 2026-05-19
[Question] Support for CLI-based executors (claude-cli) vs. Direct API Integration


qitma:
I am exploring the integration of Evonic with claude-cli to execute headless prompts. Specifically, I am curious about the intended flow for Super Agents:

Is it possible to have Evonic trigger a local CLI command (e.g., claude-cli -p "command") and capture the output directly as an agentic action?
Or, does Evonic require a direct LLM API key integration by design, where the CLI would instead be defined as a "Skill" or tool for the agent to call?
I am trying to determine if I can bypass direct API costs by leveraging a locally authenticated CLI session.

`Evonic --> super agent --> trigger claude-cli -p "some comand" --> return output`

Thanks mas robin.

binsarjr:
Hi @qitma 👋

In my view Evonic isn't really ready to take on first-class integration with CLI-based AI tools (`claude-cli` and friends) yet. The main reason is **maintenance cost** — these CLIs change fast and frequently, and wiring one in as an executor means committing to chase every breaking change in their flags, auth flow, and output format. That's a lot of surface area to keep stable over time.

For your specific goal — avoiding direct API costs by reusing an authenticated session — there are a couple of practical paths today:

1. **OpenAI-compatible proxy** — expose your Claude subscription behind an OpenAI-compatible endpoint using a router/proxy like [9router](https://github.com/decolua/9router), then point Evonic at that endpoint as a regular provider. That's the setup I've been running myself.
2. **Skill + dedicated agent** — define `claude -p` as a Skill/tool and wire it to a dedicated agent whose whole purpose is to always invoke that `claude -p` call. So instead of Evonic owning the CLI integration, the agent is instructed to route everything through `claude -p`.

Just go in with eyes open on either path: they carry their own trade-offs (reliability, and it sits outside the officially supported usage), so weigh the risks before relying on them.

More broadly, I think Evonic would benefit from **solidifying its core fundamentals first** before expanding into a lot of external tool integrations. Otherwise the integration surface tends to grow faster than it can realistically be maintained. Honestly this is partly just a personal concern of mine — a lot of the other agentic swarms out there feel way too bloated, and I'd hate to see Evonic head down that same path.

That said, this is just my take — curious how others see it. Do you think CLI executors are worth the maintenance overhead, or would an API/proxy approach cover most use cases? Happy to discuss further.

anvie:
[...]

PR #100, opened 2026-08-06
feat: add Claude Code auth and improve Codex setup

binsarjr:
## Context

Evonic already had a working Codex transport and a basic OAuth callback flow. This PR does not replace that implementation. It fills in the missing credential-management and setup experience around it, then adds the same kind of first-class support for Claude Code / Anthropic.

For Codex, the existing transport and legacy routes are kept intact. The new pieces are:

- detection of an existing Codex CLI login from CODEX_HOME/auth.json or the macOS Keychain
- an explicit option to import that credential snapshot into Evonic
- a separate OpenAI device-code login for users who do not want to reuse their CLI session
- safer provider lookup so another OAuth provider cannot accidentally be treated as Codex
- refresh handling that checks for a newer CLI credential before spending a possibly stale refresh token

## Claude Code / Anthropic support

This adds a native Anthropic provider with three setup choices:

- use the current Claude Code login from the macOS Keychain or ~/.claude/.credentials.json
- sign in with a separate Claude Pro / Max account through PKCE OAuth
- provide an Anthropic API key or a token generated by claude setup-token

When the Claude Code credential store is selected, Evonic reads it directly instead of copying the token into the providers table. OAuth requests use the Claude Code headers and system identity expected by the Anthropic Messages API. Refresh tokens are re-read before refresh and file updates are written atomically with mode 0600.

The credential behavior follows the same approach used by Hermes Agent, while staying within Evonic's existing provider and LLM client structure.

## Settings experience

Codex and Anthropic now appear as built-in providers on both new and existing installations. Their cards show whether setup is needed and open a dedicated setup dialog with the available credential choices. The dialog also supports disconnecting Evonic without deleting external CLI credentials.

Provider API responses now expose only a configured/not-configured flag; access and refresh tokens are removed before the response reaches the browser. The dialog was checked at desktop and mobile widths, including keyboard focus, error messages, and overflow behavior.

## Compatibility

- existing Codex routes remain available
- existing Codex provider rows are migrated to the new credential-source field
- model-level API key overrides still take precedence
- no new dependency was added

## Verification

- 15 focused provider and Codex tests passed locally
- Python compilation, JavaScript syntax, and git diff checks passed
- Go CI passed
- Python CI completed with 1998 passed and 92 skipped; its only failure is the existing SFTP Mock context-manager test, which also reproduces on a clean origin/dev checkout

This is intentionally a draft while the credential setup flows are tested manually in the running UI.

### DIFF
--- a/app.py
+++ b/app.py
@@ -93,6 +93,7 @@ from routes.scheduler import scheduler_bp
 from routes.models import models_bp
 from routes.providers import providers_bp
 from routes.codex import codex_bp
+from routes.claude import claude_bp
 from routes.health import health_bp
 from routes.workplaces import workplaces_bp
 from routes.logs import logs_bp
@@ -240,6 +241,7 @@ app.register_blueprint(dashboard_bp)
 app.register_blueprint(models_bp)
 app.register_blueprint(providers_bp)
 app.register_blueprint(codex_bp)
+app.register_blueprint(claude_bp)
 app.register_blueprint(health_bp)
 app.register_blueprint(workplaces_bp)
 app.register_blueprint(logs_bp)
--- a/backend/llm_client.py
+++ b/backend/llm_client.py
@@ -212,7 +212,9 @@ class LLMClient:
                          If None, uses the default model from DB or config.py defaults.
         """
         self.provider = None
+        self._model_api_key_override = False
         if model_config:
+            self._model_api_key_override = bool(model_config.get("api_key"))
             try:
                 from models.db import db
                 model_config = db.resolve_model_config(model_config)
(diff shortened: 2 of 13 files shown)


binsarjr:
Related: #45.

This is **BYOR**: use your own account, at your own risk. Credential reuse is opt-in; Evonic does not share accounts or bypass provider limits.

Provider terms can change, and misuse may lead to token revocation or an account ban. If a flow is no longer allowed, we should disable it instead of working around enforcement. Use official API credentials for production.

References: [Anthropic credential policy](https://code.claude.com/docs/en/legal-and-compliance#authentication-and-credential-use) · [OpenAI Terms](https://openai.com/policies/terms-of-use/)

binsarjr:
Superseded by #104. The same provider change was cherry-picked onto the dedicated `binsarjr:feature/codex-claude-code-providers` branch based directly on `anvie/dev`, so future aggregation commits on `binsarjr/dev` cannot leak into the PR.

PR #102, opened 2026-08-07
feat: per-agent daily token limit enforcement (fails open)

fepfitra:
Per-agent daily token cap for LLM usage, using the existing token_monitor telemetry.

- handle_message blocks turns when the agent's today-token sum >= limit (super agents exempt)
- Config: per-agent `token_limit_daily` column (NULL = env default `EVONIC_TOKEN_LIMIT_DAILY_DEFAULT`, default 500k tokens/day)
- Emits `limit_hit` event on block; fails open on any limiter error (never breaks the agent path)
- Reads the existing token_monitor `token_usage` table — no new storage

Verified: token_monitor tests pass (4), block logic tested end-to-end (over/under/super/custom limit), schema migration applies at startup.


anvie:
Can you please fix tho conflict first?

PR #104, opened 2026-08-10
feat: add Claude Code auth and improve Codex setup

binsarjr:
## Branch hygiene

This replaces #100 with the same provider commit cherry-picked onto a dedicated branch based directly on `anvie/dev`. It prevents unrelated future commits on `binsarjr/dev` from entering the PR.

## Context

Evonic already had a working Codex transport and a basic OAuth callback flow. This PR does not replace that implementation. It fills in the missing credential-management and setup experience around it, then adds the same kind of first-class support for Claude Code / Anthropic.

For Codex, the existing transport and legacy routes are kept intact. The new pieces are:

- detection of an existing Codex CLI login from CODEX_HOME/auth.json or the macOS Keychain
- an explicit option to import that credential snapshot into Evonic
- a separate OpenAI device-code login for users who do not want to reuse their CLI session
- safer provider lookup so another OAuth provider cannot accidentally be treated as Codex
- refresh handling that checks for a newer CLI credential before spending a possibly stale refresh token

## Claude Code / Anthropic support

This adds a native Anthropic provider with three setup choices:

- use the current Claude Code login from the macOS Keychain or ~/.claude/.credentials.json
- sign in with a separate Claude Pro / Max account through PKCE OAuth
- provide an Anthropic API key or a token generated by claude setup-token

When the Claude Code credential store is selected, Evonic reads it directly instead of copying the token into the providers table. OAuth requests use the Claude Code headers and system identity expected by the Anthropic Messages API. Refresh tokens are re-read before refresh and file updates are written atomically with mode 0600.

The credential behavior follows the same approach used by Hermes Agent, while staying within Evonic's existing provider and LLM client structure.

## Settings experience

Codex and Anthropic appear as built-in providers. Provider API responses expose only a configured/not-configured flag; access and refresh tokens are removed before responses reach the browser.

## Compatibility

- existing Codex routes remain available
- existing Codex provider rows are migrated to the new credential-source field
- model-level API key overrides still take precedence
- no new dependency was added

## Verification

- the original focused provider suite reported 15 passing tests
- Python compilation and git diff checks pass after the clean cherry-pick
- the resulting branch contains one feature commit and only the original 13 files

### DIFF
--- a/app.py
+++ b/app.py
@@ -93,6 +93,7 @@ from routes.scheduler import scheduler_bp
 from routes.models import models_bp
 from routes.providers import providers_bp
 from routes.codex import codex_bp
+from routes.claude import claude_bp
 from routes.health import health_bp
 from routes.workplaces import workplaces_bp
 from routes.logs import logs_bp
@@ -240,6 +241,7 @@ app.register_blueprint(dashboard_bp)
 app.register_blueprint(models_bp)
 app.register_blueprint(providers_bp)
 app.register_blueprint(codex_bp)
+app.register_blueprint(claude_bp)
 app.register_blueprint(health_bp)
 app.register_blueprint(workplaces_bp)
 app.register_blueprint(logs_bp)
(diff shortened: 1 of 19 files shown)


binsarjr:
Fast mode added in d9047083. Verification: 39 focused tests and 105 Kanban plugin tests pass locally; all 6 new fast-mode tests also pass in CI. Python CI remains red only on the two existing baseline failures test_ssh_backend_binary_read and test_token_efficiency, which also fail on current anvie/dev (run 31512574718). Go tests pass.

PR #107, opened 2026-08-13
feat(pdf): add native PDF attachment analysis

binsarjr:
## Summary

- add `analyze_pdf`, which sends owned PDF attachments to provider-native document inputs for Gemini, OpenAI-compatible, and Anthropic models
- add a model capability flag, global primary plus two PDF fallbacks, and a per-agent PDF Analysis toggle with a managed tool assignment
- keep PDF bytes out of the main conversation prompt; web and Telegram expose only attachment metadata and the tool hint
- remove eager `pypdf` extraction and add a Google Gemini provider preset without adding dependencies

## Safety and behavior

- validate attachment ID, agent ownership, session ownership, PDF signature, query length, and provider-specific size limits
- treat PDF contents as untrusted data and tell the analysis model not to follow instructions embedded in the document
- try configured PDF models in order, then other enabled PDF-capable models

## Verification

- focused native-PDF and integration suites: 76 passed
- final regression subset after rebase: 36 passed
- full unit suite: 2141 passed, 85 skipped, 3 pre-existing failures reproduced on `origin/dev`
  - `test_credential_extraction_wont_block_at_default`
  - `test_cat_file_bytes_reads_binary_data_through_sftp`
  - `test_artifact_instructions_keep_capabilities_with_lower_token_cost`
- Python compile, JavaScript syntax checks, and `git diff --check` passed

Gemini/OpenAI/Anthropic HTTP payloads are covered with mocked contract tests. A live Gemini call was not run because this environment has no configured Gemini credential.

This branch is one commit directly on `origin/dev`, does not depend on PR #104, and has not been installed into the live `~/.evonic` checkout.

### DIFF
--- a/backend/agent_portability.py
+++ b/backend/agent_portability.py
@@ -15,7 +15,7 @@ VARIABLE_KEY_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
 MAX_TEXT_LENGTH = 102400
 MAX_KB_FILE_LENGTH = 1024 * 1024
 CONFIG_KEYS = frozenset({
-    "enabled", "vision_enabled", "summarize_threshold", "summarize_tail",
+    "enabled", "vision_enabled", "pdf_enabled", "summarize_threshold", "summarize_tail",
     "summarize_prompt", "message_buffer_seconds", "inject_agent_id",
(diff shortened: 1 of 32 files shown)


binsarjr:
Continued in #108. Closing this draft so review stays in one place.

